PR #10, opened 2022-08-11
Rename compile env name for gnmi write

ganglyu:
Signed-off-by: Gang Lv ganglv@microsoft.com

PR for sonic-buildimage
https://github.com/sonic-net/sonic-buildimage/pull/11780

#### Why I did it
I will add a future env var ENABLE_NATIVE_WRITE which is another kind of gnmi write mode.
So, I need to rename TELEMETRY_WRITABLE and READ_WRITE_MODE to avoid conflict.

#### How I did it
Rename compile env var and build tag.

#### How to verify it
Run unit test for sonic-gnmi.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,8 +17,8 @@ SRC_FILES=$(shell find . -name '*.go' | grep -v '_test.go' | grep -v '/tests/')
 TEST_FILES=$(wildcard *_test.go)
 TELEMETRY_TEST_DIR = build/tests/gnmi_server
 TELEMETRY_TEST_BIN = $(TELEMETRY_TEST_DIR)/server.test
-ifeq ($(TRANSLIB_ENABLE),y)
-BLD_FLAGS := -tags translib
+ifeq ($(TRANSLIB_ENABLED),y)
+BLD_FLAGS := -tags gnmi_translib
 endif
 
 GO_DEPS := vendor/.done
--- a/Makefile
+++ b/Makefile
@@ -17,8 +17,8 @@ SRC_FILES=$(shell find . -name '*.go' | grep -v '_test.go' | grep -v '/tests/')
 TEST_FILES=$(wildcard *_test.go)
 TELEMETRY_TEST_DIR = build/tests/gnmi_server
 TELEMETRY_TEST_BIN = $(TELEMETRY_TEST_DIR)/server.test
-ifeq ($(TELEMETRY_WRITABLE),y)
-BLD_FLAGS := -tags readwrite
+ifeq ($(TRANSLIB_ENABLE),y)
+BLD_FLAGS := -tags translib
 endif
 
 GO_DEPS := vendor/.done
--- a/Makefile
+++ b/Makefile
@@ -17,8 +17,8 @@ SRC_FILES=$(shell find . -name '*.go' | grep -v '_test.go' | grep -v '/tests/')
 TEST_FILES=$(wildcard *_test.go)
 TELEMETRY_TEST_DIR = build/tests/gnmi_server
 TELEMETRY_TEST_BIN = $(TELEMETRY_TEST_DIR)/server.test
-ifeq ($(TRANSLIB_ENABLED),y)
-BLD_FLAGS := -tags gnmi_translib
+ifeq ($(ENABLE_TRANSLIB_WRITE),y)
+BLD_FLAGS := -tags gnmi_translib_write
 endif
 
 GO_DEPS := vendor/.done
--- a/gnmi_server/server.go
+++ b/gnmi_server/server.go
@@ -48,9 +48,8 @@ type Config struct {
 	// for this Server.
 	Port     int64
 	UserAuth AuthTypes
-	MgmtEnable      bool
+	TranslibEnable  bool
 	TelemetryEnable bool
-	MixedEnable     bool
 }
 
 var AuthLock sync.Mutex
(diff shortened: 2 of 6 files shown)


gnmi_server/server.go
```diff
@@ -140,11 +143,11 @@ func NewServer(config *Config, opts []grpc.ServerOption) (*Server, error) {
 	}
 	gnmipb.RegisterGNMIServer(srv.s, srv)
 	spb_jwt_gnoi.RegisterSonicJwtServiceServer(srv.s, srv)
-	if READ_WRITE_MODE {
```
qiluo-msft, line 143: >READ_WRITE_MODE [](http://example.com/codeflow?start=4&length=15)

This is defined in gnmi_server/constants.go. Will you remove it completely.

gnmi_server/server.go
```diff
@@ -140,11 +143,11 @@ func NewServer(config *Config, opts []grpc.ServerOption) (*Server, error) {
...
+	if srv.config.MgmtEnable {
```
qiluo-msft, line 146: >MgmtEnable [](http://example.com/codeflow?start=15&length=10)

Suggest only refactor in this PR:
1. keep original naming: `srv.config.ReadWriteMode`
2. I guess your new feature is `MixedEnable`. Let's add it later in another PR.

gnmi_server/server.go
```diff
@@ -230,6 +233,9 @@ func (s *Server) Subscribe(stream gnmipb.GNMI_SubscribeServer) error {
 		return nil, status.Error(codes.PermissionDenied, msg)
 	}
 	*/
+	if !s.config.TelemetryEnable {
```
qiluo-msft, line 236: >TelemetryEnable [](http://example.com/codeflow?start=14&length=15)

I am not clear about the necessity for disabling subscribe. Is there a use case of subscribing a path in ConfigDB? if yes, no need to add option.

Makefile
```diff
@@ -17,8 +17,8 @@ SRC_FILES=$(shell find . -name '*.go' | grep -v '_test.go' | grep -v '/tests/')
 TEST_FILES=$(wildcard *_test.go)
 TELEMETRY_TEST_DIR = build/tests/gnmi_server
 TELEMETRY_TEST_BIN = $(TELEMETRY_TEST_DIR)/server.test
-ifeq ($(TELEMETRY_WRITABLE),y)
-BLD_FLAGS := -tags readwrite
+ifeq ($(TRANSLIB_ENABLE),y)
+BLD_FLAGS := -tags translib
```
qiluo-msft, line 21: >translib [](http://example.com/codeflow?start=19&length=8)

I propose `gnmi_translib` and anticipate more flags like `gnmi_native`, etc.

qiluo-msft: @seiferteric Could you help check the name?

Makefile
```diff
@@ -17,8 +17,8 @@ SRC_FILES=$(shell find . -name '*.go' | grep -v '_test.go' | grep -v '/tests/')
 TEST_FILES=$(wildcard *_test.go)
 TELEMETRY_TEST_DIR = build/tests/gnmi_server
 TELEMETRY_TEST_BIN = $(TELEMETRY_TEST_DIR)/server.test
-ifeq ($(TELEMETRY_WRITABLE),y)
-BLD_FLAGS := -tags readwrite
+ifeq ($(TRANSLIB_ENABLE),y)
```
qiluo-msft, line 20: >TRANSLIB_ENABLE [](http://example.com/codeflow?start=8&length=15)

SOMETHING_ENABLED
or ENABLE_SOMETHING

qiluo-msft (approved):

Makefile
```diff
@@ -17,8 +17,8 @@ SRC_FILES=$(shell find . -name '*.go' | grep -v '_test.go' | grep -v '/tests/')
 TEST_FILES=$(wildcard *_test.go)
 TELEMETRY_TEST_DIR = build/tests/gnmi_server
 TELEMETRY_TEST_BIN = $(TELEMETRY_TEST_DIR)/server.test
-ifeq ($(TELEMETRY_WRITABLE),y)
-BLD_FLAGS := -tags readwrite
+ifeq ($(TRANSLIB_ENABLED),y)
```
qiluo-msft, line 20: >TRANSLIB_ENABLED [](http://example.com/codeflow?start=8&length=16)

Since you rename it, will you raise a buildimage PR? Mention the related PRs in this PR description.

seiferteric:
Can I get more context on this? Before we had a build flag to make gNMI read-only, with this there will be a build flag and a command line option? Is the intention to have both? Is the build flag still being used? What i the build flag is set to false, but the command line set to true? Or is the intent to remove the build flag and make it a run-time command line option? I thought it was made a build flag because there was a request to be able to have be able to compile gNMI without the ability to write to the DB? Also you get memory savings with the build flag as the mgmt-framework is compiled out.

Also, the new option is called gnmi_translib but should it not still be something like gnmi_readonly? I guess translib data client is the only one supporting write ability for now, but shouldn't we keep the option name more generic to indicate the intention? Why change the name to gnmi_translib?

ganglyu:
> Can I get more context on this? Before we had a build flag to make gNMI read-only, with this there will be a build flag and a command line option? Is the intention to have both? Is the build flag still being used? What i the build flag is set to false, but the command line set to true? Or is the intent to remove the build flag and make it a run-time command line option? I thought it was made a build flag because there was a request to be able to have be able to compile gNMI without the ability to write to the DB? Also you get memory savings with the build flag as the mgmt-framework is compiled out.
> 
> Also, the new option is called gnmi_translib but should it not still be something like gnmi_readonly? I guess translib data client is the only one supporting write ability for now, but shouldn't we keep the option name more generic to indicate the intention? Why change the name to gnmi_translib? Also with this change translib is still enabled for Get/Subscribe operations correct?

The intention is to have a build flag and a command line option, and it will be more flexible, for example, build flag is disabled, and we still want to run related unit test. Now build flag decides the default value of the command line option, so if the build flag is set to false, but the command line set to true, this feature will be enabled. But we will not use command line in this way.

We will introduce other gnmi write feature, so we want to use two names: gnmi_translib and gnmi_native.
Since translib is still enabled for Get/Subscribe operations, maybe we can use gnmi_translib_write and gnmi_native_write?

qiluo-msft:
"xxx_write" may be super accurate since that feature support GetRequest and SetRequest.

ganglyu:
@seiferteric
I have updated the name, would you please review again?

qiluo-msft (approved):

PR #38, opened 2022-09-01
Improve unit test for gnmi

ganglyu:
Signed-off-by: Gang Lv ganglv@microsoft.com

#### Why I did it
We should not run unit test without redis server.

#### How I did it
Check redis.sock, if there's no redis.sock, do not run unit test.

#### How to verify it
Run unit test for gnmi, and build image.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ else
 	$(GO) install -mod=vendor github.com/sonic-net/sonic-gnmi/gnoi_client
 endif
 
-ut:
+check_gotest:
 	sudo mkdir -p ${DBDIR}
 	sudo cp ./testdata/database_config.json ${DBDIR}
 	sudo mkdir -p /usr/models/yang || true
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,6 @@ endif
 GO_DEPS := vendor/.done
 PATCHES := $(wildcard patches/*.patch)
 PATCHES += $(shell find $(MGMT_COMMON_DIR)/patches -type f)
-REDIS_SOCK := $(shell if sudo [ -e /var/run/redis/redis.sock ]; then echo "exist"; else echo "noexist"; fi)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -77,6 +79,9 @@ check:
 	$(GO) get github.com/AlekSi/gocov-xml
 	gocov convert coverage-*.txt | gocov-xml -source $(shell pwd) > coverage.xml
 	rm -rf coverage-*.txt 
+else
+	$(info Please start redis server to run unit test)
```
qiluo-msft, line 83: This will be treated as test passed? Is it expected?

Makefile
```diff
@@ -24,6 +24,7 @@ endif
 GO_DEPS := vendor/.done
 PATCHES := $(wildcard patches/*.patch)
 PATCHES += $(shell find $(MGMT_COMMON_DIR)/patches -type f)
+REDIS_SOCK := $(shell if sudo [ -e /var/run/redis/redis.sock ]; then echo "exist"; else echo "noexist"; fi)
```
qiluo-msft, line 27: Instead of checking a file, how about `redis-cli PING`?

Makefile
ganglyu: When we create a PR for sonic-gnmi repo, azure pipeline will start redis server, and we can run unit test.
But when we use sonic-buildimage to build, there's no redis server, and we should not run unit test.

Makefile
```diff
@@ -65,7 +65,7 @@ else
 	$(GO) install -mod=vendor github.com/sonic-net/sonic-gnmi/gnoi_client
 endif
 
-check:
+ut:
```
qiluo-msft, line 68: >ut [](http://example.com/codeflow?start=0&length=2)

`check_gotest` ?

qiluo-msft (approved):

PR #45, opened 2022-10-12
Streaming events URL support "not to use cache"

renukamanavalan:
#### Why I did it
Only one gNMI client can use cache at any time. 
The external clients that stream events reliably need to use cache.
While an external client is connected, if we need to use some test client (e.g. gnmi_cli), it could be used provided it does not use cache.

#### How I did it
Added support for a new URL parameter [usecache=false].  This parameter will *disable* use of cache.

#### How to verify it
1) Have no gNMI client
2) Ensure some events are published (may use events_tool to manually publish)
3) Run gnmi_cli with [usecache=false]
4) Note, it does not receive/spew any events from past.
5) Publish some events (may use events_tool)
6) See the client receive these
7) Now run a client with cache use *not* disabled
8) Note, it receives all events including those from step 2

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,31 @@ You can also build a debian package and install it:
 * See [SONiC telemetry in dial-out mode](./doc/dialout.md) for how to run dial-out mode system telemetry client
 * See [gNMI Usage Examples](./doc/gNMI_usage_examples.md) for gNMI client usage examples.
 
+### Streaming events
+SONiC publishes events of interest via gNMI.
+Sample events could be bgp-state change, link-state change.
+A sample query URL is as below.
+    `gnmi_cli -client_types=gnmi -a 127.0.0.1:50051 -t EVENTS -logtostderr -insecure -v 7 -streaming_type ON_CHANGE -q all -qt s`
+
+The events URL can take the following optional parameters.<br/>
+heartbeat:<br/>
+    `[heartbeat=<N>]`
+<br/>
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -39,6 +39,31 @@ You can also build a debian package and install it:
...
+<br/>
+&nbsp;&nbsp;&nbsp;&nbsp;The SONiC switch publishes periodic hearbeats when there are no events to publish.<br/>
+&nbsp;&nbsp;&nbsp;&nbsp;The frequency of the heartbeat can be controlled by this parameter as beat in every N seconds.<br/>
+
+usecache:<br/>
+    `[usecache=true/false]`
+<br/>
+&nbsp;&nbsp;&nbsp;&nbsp;The SONiC switch does offline cache when gNMI client is down.<br/>
+&nbsp;&nbsp;&nbsp;&nbsp;The cached events are delivered upon nest gNMI connection<br/>
+&nbsp;&nbsp;&nbsp;&nbsp;Only one client should use the cache service<br/>
```
ganglyu, line 60: How to guarantee this limitation?
If two clients use the cache service, what will happen?

renukamanavalan: Limitation by design. 
Trade off between code complexity in supporting multiple clients with cache vs practical use case.
In practice, there is only one client streaming out.
This is added to enable test tools to function w/o disturbing the main service.
To answer your question: If multiple use cache, of course first come first served. The cache will go to one or other.

ganglyu (approved):

PR #47, opened 2022-10-21
The dialout client support the source ip

rockyrth:
Why I did it
Each device has many ip addresses， but not all of them are published to network.
Default, the dialout_client_cli just select one of them as source ip, but the dialout_server may not know the source ip from dialout_client.
So we know to select the sepcial source ip for dialout_client_cli

How I did it
Use grpc.WithContextDialer to add a function which support dialout with source ip

How to verify it
CONFIG_DB HSET TELEMETRY_CLIENT|Global src_ip x.x.x.x
then, the dialout_server will receive the src_ip above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,12 +34,12 @@ You can also build a debian package and install it:
     dpkg-buildpackage -rfakeroot -b -us -uc
     popd
 
-### Running
+## Running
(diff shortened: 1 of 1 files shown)


qiluo-msft:
Could you check the test failure?

ganglyu:
Please add unit test for this change.

Issue #42, opened 2022-09-23
Unordered map serialization, cause different results of the same operation


panyi5534:
When I use gnmi_get to get data, if the path contains 2 or more keys, the result of the same operation is different.

As I know, it is caused by the unordered map of proto serialization. For example, the path "/sonic-acl:sonic-acl/ACL_RULE/ACL_RULE_LIST[aclname=1][rulename=2]" convert to [sonic-acl:sonic-aclACL_RULE ACL_RULE_LIST map[aclname:1 rulename:2]]，but the map is not ordered,  so the result of proto serialization cannot be accurate.

Operation record:
```
root@sonic:/# gnmi_get -xpath_target SONiC-YANG -xpath /sonic-acl:sonic-acl/ACL_RULE/ACL_RULE_LIST[aclname=1][rulename=2] -target_addr 127.0.0.1:8090 -insecure true
== getRequest:
prefix: <
  target: "SONiC-YANG"
>
path: <
  elem: <
    name: "sonic-acl:sonic-acl"
  >
  elem: <
    name: "ACL_RULE"
  >
  elem: <
    name: "ACL_RULE_LIST"
    key: <
      key: "aclname"
      value: "1"
    >
    key: <
      key: "rulename"
      value: "2"
    >
  >
>
encoding: JSON_IETF

== getResponse:
notification: <
  timestamp: 1663901622269512774
  prefix: <
    target: "SONiC-YANG"
  >
  update: <
    path: <
      elem: <
        name: "sonic-acl:sonic-acl"
      >
      elem: <
        name: "ACL_RULE"
      >
      elem: <
        name: "ACL_RULE_LIST"
        key: <
          key: "aclname"
          value: "1"
        >
        key: <
          key: "rulename"
          value: "2"
        >
      >
    >
    val: <
      json_ietf_val: "{\"sonic-acl:ACL_RULE_LIST\":[{\"DST_IP\":\"192.168.110.5\",\"IP_TYPE\":\"IP\",\"PACKET_ACTION\":\"REDIRECT\",\"SRC_IP\":\"192.168.110.4\",\"aclname\":\"1\",\"rulename\":\"2\"}]}"
    >
  >
>

root@sonic:/# gnmi_get -xpath_target SONiC-YANG -xpath /sonic-acl:sonic-acl/ACL_RULE/ACL_RULE_LIST[aclname=1][rulename=2] -target_addr 127.0.0.1:8090 -insecure true
== getRequest:
prefix: <
[...]

PR #50, opened 2022-10-25
fixes #issue 42. add num for each key and sort keys by num. make sure keys in map can be sorted.

pettershao-ragilenetworks:
#### Why I did it
The path is not fixed due to the unordered property of the map，which may cause the resource to be unable to be found.
<https://github.com/sonic-net/sonic-gnmi/issues/42>

#### How I did it
add num for each key and sort keys by num.

#### How to verify it
gnmi_get,gnmi_set's path has more than two keys.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/patches/xpath.patch
+++ b/patches/xpath.patch
@@ -1,0 +1,34 @@
+--- ./github.com/jipanyang/gnxi/utils/xpath/xpath.go	2022-10-25 16:16:00.239783329 +0800
++++ ./github.com/jipanyang/gnxi/utils/xpath/xpath.go	2022-10-25 15:56:18.336168598 +0800
+@@ -20,6 +20,7 @@
+ 	"fmt"
+ 	"regexp"
+ 	"strings"
++	"strconv"
+ )
+ 
+ var (
+@@ -135,9 +136,10 @@
+ // For example, given
+ //	"[k1=v1][k2=v2]",
+ // this API returns
+-//	map[string]string{"k1": "v1", "k2": "v2"}.
++//	map[string]string{"[0]k1": "v1", "[1]k2": "v2"}.
+ func parseKeyValueString(str string) (map[string]string, error) {
+ 	keyValuePairs := make(map[string]string)
++	var num int = 0
+ 	// begin marks the beginning of a key-value pair.
+ 	begin := 0
+ 	// end marks the end of a key-value pair.
+@@ -170,7 +172,10 @@
+ 				// Recover escaped '[' and ']'.
+ 				val = strings.Replace(val, `\]`, `]`, -1)
+ 				val = strings.Replace(val, `\[`, `[`, -1)
+-				keyValuePairs[key] = val
++//				keyValuePairs[key] = val
++				key_with_num := "[" + strconv.Itoa(num) + "]" + key
++				keyValuePairs[key_with_num] = val
++				num++
+ 				begin = end + 1
+ 			}
+ 			end++
--- a/transl_utils/transl_utils.go
+++ b/transl_utils/transl_utils.go
@@ -12,7 +12,8 @@ import (
 	"context"
 	"log/syslog"
 	"github.com/Azure/sonic-mgmt-common/translib/tlerr"
-
+	"strconv"
+	"sort"
 )
 
 var (
@@ -80,6 +81,9 @@ func ConvertToURI(prefix *gnmipb.Path, path *gnmipb.Path, req *string) error {
 	elems := fullPath.GetElem()
 	*req = "/"
 
+	num_key_map := make(map[int]string)
+	key_val_map := make(map[string]string)
+
 	if elems != nil {
 		/* Iterate through elements. */
 		for i, elem := range elems {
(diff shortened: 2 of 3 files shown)


patches/xpath.patch
```diff
@@ -0,0 +1,34 @@
...
+ func parseKeyValueString(str string) (map[string]string, error) {
+ 	keyValuePairs := make(map[string]string)
++	var num int = 0
+ 	// begin marks the beginning of a key-value pair.
+ 	begin := 0
+ 	// end marks the end of a key-value pair.
+@@ -170,7 +172,10 @@
+ 				// Recover escaped '[' and ']'.
+ 				val = strings.Replace(val, `\]`, `]`, -1)
+ 				val = strings.Replace(val, `\[`, `[`, -1)
```
renukamanavalan, line 26: Good find!

transl_utils/transl_utils.go
```diff
@@ -94,8 +98,32 @@ func ConvertToURI(prefix *gnmipb.Path, path *gnmipb.Path, req *string) error {
...
+					    return err
+					}
+					num_key_map[num] = key_del_num
+					key_val_map[key_del_num] = v
+				}
+
+				log.V(6).Infof("num_key_map : %#v", num_key_map)
+				log.V(6).Infof("key_val_map : %#v", key_val_map)
+
+				if num_key_map != nil {
```
renukamanavalan, line 118: Always return != nil, as you have made it with make. You can check len instead.

transl_utils/transl_utils.go
```diff
@@ -94,8 +98,32 @@ func ConvertToURI(prefix *gnmipb.Path, path *gnmipb.Path, req *string) error {
 			/* If keys are present , process the keys. */
 			if key != nil {
 				for k, v := range key {
-					log.V(6).Infof("elem : %#v %#v", k, v)
-					*req += "[" + k + "=" + v + "]"
+					//log.V(6).Infof("elem : %#v %#v", k, v)
```
renukamanavalan, line 101: Please remove commented lines.

transl_utils/transl_utils.go
```diff
@@ -94,8 +98,32 @@ func ConvertToURI(prefix *gnmipb.Path, path *gnmipb.Path, req *string) error {
...
+					//*req += "[" + k + "=" + v + "]"
+					num_str := k[strings.Index(k, "[") + 1 : strings.Index(k, "]")]
+					log.V(6).Infof("num : %#v", num_str)
+					key_del_num := k[strings.Index(k, "]") + 1 : len(k)]
+					log.V(6).Infof("elem : %#v %#v", key_del_num, v)
+					num, err := strconv.Atoi(num_str)
```
renukamanavalan, line 107: You may remove these 2 log lines. Later log at 115 & 116 should be sufficient.

transl_utils/transl_utils.go
```diff
@@ -94,8 +98,32 @@ func ConvertToURI(prefix *gnmipb.Path, path *gnmipb.Path, req *string) error {
...
+
+				log.V(6).Infof("num_key_map : %#v", num_key_map)
+				log.V(6).Infof("key_val_map : %#v", key_val_map)
+
+				if num_key_map != nil {
+					var num_list []int
+					for num_key_map_k := range num_key_map {
+					    num_list = append(num_list, num_key_map_k)
+					}
+					sort.Ints(num_list)
```
renukamanavalan, line 123: Shouldn't this be already sorted?
xPath patch uses it in sequence.

pettershao-ragilenetworks:
@renukamanavalan I fixed by suggestion fix, and add a cace when key has no tag for test module.

renukamanavalan (approved):

renukamanavalan:
@ganglyu & @zbud-msft  - Please provide your review/approval.

patches/xpath.patch
```diff
@@ -0,0 +1,34 @@
...
++	"strconv"
+ )
+ 
+ var (
+@@ -135,9 +136,10 @@
+ // For example, given
+ //	"[k1=v1][k2=v2]",
+ // this API returns
+-//	map[string]string{"k1": "v1", "k2": "v2"}.
++//	map[string]string{"[0]k1": "v1", "[1]k2": "v2"}.
```
ganglyu, line 16: How about compatibility?
If we add num for each key, other GNMI client can't communicate with SONiC telemetry?

renukamanavalan: The scope of this numbering within parameters of a path element.

@pettershao-ragilenetworks  -- please correct my statement if needed.

ganglyu:
> this bug cause by GNMI client, the client assign value to map with "keyValuePairs[key] = val", causes the k-vs ​​to be out of order in memory. I used the int(-9223372036854775808 ~ 9223372036854775807) type in c-s, but there are rarely more than 10 keys in practice. If key without tag, the request also can be accepted, but the order of map to path cannot be guaranteed, and the resource may not be found. see file transl_utils/transl_utils.go in line 101. @ganglyu & @zbud-msft - Please provide your review/approval.

I wonder if GNMI spec has any solution for this order issue.

ganglyu:
https://github.com/openconfig/gnmi/blob/master/proto/gnmi/gnmi.proto#L153-L156
GNMI protobuf is using map for key, and map is ordered by key. That's why we can't get the expected order.
Translib is used to support openconfig yang, maybe we can restore the order from yang model?

pettershao-ragilenetworks:
I had try use list[map,map] struct replace map or add a list to record key in protobuf(modify protobuf defined message), but failed. Pkg YGOT relies on the structure generated by protobuf, cause can not use YGOT, so at last I gived up, and the code is more than now. @ganglyu

hdwhdw:
Closing this PR as it has been inactive for two years. Feel free to reopen if needed.

PR #55, opened 2022-11-03
Add conditional check for split

zbud-msft:
#### Why I did it

Can crash if no delimitor is present and accessing second part of split

Crash stack
```
telemetry#/supervisord: telemetry #011/sonic/src/sonic-telemetry/sonic_data_client/db_client.go:
telemetry#/supervisord: telemetry #011/sonic/src/sonic-telemetry/sonic_data_client/db_client.go:315 +0x14f
telemetry#/supervisord: telemetry #011/sonic/src/sonic-telemetry/sonic_data_client/db_client.go:653 +0xcb2
telemetry#/supervisord: telemetry #011/sonic/src/sonic-telemetry/gnmi_server/server.go:317 +0x4d3
telemetry#/supervisord: telemetry #011/sonic/src/sonic-telemetry/vendor/google.golang.org/grpc/server.go:1533 +0xcfd
telemetry#/supervisord: telemetry google.golang.org/grpc.(*Server).serveStreams.func1.2(0xc0013e2640, 0xc00094ba40, 0x11758e0, 0xc002222480, 0xc000e6b100)
telemetry#/supervisord: telemetry github.com/Azure/sonic-telemetry/sonic_data_client.(*DbClient).Get(0xc000182e70, 0x0, 0x1, 0xc0000108d0, 0x1171b60, 0xc000182e70, 0x0)
telemetry#/supervisord: telemetry #011/sonic/src/sonic-telemetry/vendor/github.com/openconfig/gnmi/proto/gnmi/gnmi.pb.go:3381 +0x217
telemetry#/supervisord: telemetry #011/sonic/src/sonic-telemetry/vendor/google.golang.org/grpc/server.go:871 +0xa1
```
#### How I did it

Added conditional check

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/sonic_data_client/db_client.go
+++ b/sonic_data_client/db_client.go
@@ -719,9 +719,9 @@ func tableData2Msi(tblPath *tablePath, useKey bool, op *string, msi *map[string]
 			var key string
 			// Split dbkey string into two parts and second part is key in table
 			keys := strings.SplitN(dbkey, tblPath.delimitor, 2)
-			//if len(keys) < 2 {
-			//	return fmt.Errorf("dbkey: %s, failed split from delimitor %v", dbkey, tblPath.delimitor)
-			//}
+			if len(keys) < 2 {
+				return fmt.Errorf("dbkey: %s, failed split from delimitor %v", dbkey, tblPath.delimitor)
+			}
 			key = keys[1]
 			err = makeJSON_redis(msi, &key, op, fv)
 		}
--- a/sonic_data_client/db_client.go
+++ b/sonic_data_client/db_client.go
@@ -719,9 +719,9 @@ func tableData2Msi(tblPath *tablePath, useKey bool, op *string, msi *map[string]
 			var key string
 			// Split dbkey string into two parts and second part is key in table
 			keys := strings.SplitN(dbkey, tblPath.delimitor, 2)
-			if len(keys) < 2 {
-				return fmt.Errorf("dbkey: %s, failed split from delimitor %v", dbkey, tblPath.delimitor)
-			}
+			//if len(keys) < 2 {
+			//	return fmt.Errorf("dbkey: %s, failed split from delimitor %v", dbkey, tblPath.delimitor)
+			//}
 			key = keys[1]
 			err = makeJSON_redis(msi, &key, op, fv)
 		}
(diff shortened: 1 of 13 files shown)


ganglyu:
Can you add unit test for this case?

qiluo-msft:
Please check if we can backport this PR to release branches.

qiluo-msft:
I notice something wrong with coverage check "No lines with coverage information in this diff."
If it is working, it will block the PR for missing unit test.

ganglyu:
> sonic_data_client
https://github.com/sonic-net/sonic-gnmi/blob/master/Makefile#L74-L76
Current unit test does not count the coverage for sonic_data_client.

ganglyu:
> I notice something wrong with coverage checker "No lines with coverage information in this diff." If it is working, it will block the PR for missing unit test.

https://github.com/sonic-net/sonic-gnmi/blob/master/Makefile#L74-L76
Current unit test does not count the coverage for sonic_data_client.

sonic_data_client/db_client.go
```diff
@@ -717,6 +717,10 @@ func tableData2Msi(tblPath *tablePath, useKey bool, op *string, msi *map[string]
 			var key string
 			// Split dbkey string into two parts and second part is key in table
 			keys := strings.SplitN(dbkey, tblPath.delimitor, 2)
+			if len(keys) < 2 {
+				log.V(3).Infof("dbkey: %s, failed split from delimitor %v", dbkey, tblPath.delimitor)
```
tomek-US, line 721: This line adds to log spam. Is it _really_ necessary?

tomek-US, line 721: This line adds to log spam. Is it _really_ necessary?

zbud-msft: We don't expect to go into this if block, very rare if it does happen. If it does happen, we want to know via log

tomek-US: The problem is that if the condition persists then the log will be printed out every time this function is called. 

This function returns the same message as part of the error status. Shouldn't the caller make the decision if the error message should be printed?

zbud-msft: Yes, I will remove as we are returning error which will be handled by caller

tomek-US (approved):

ganglyu (approved):

PR #60, opened 2022-11-14
Add gnmi_dump tool for debug and unit test

ganglyu:
#### Why I did it
Add tool for debug and unit test.

#### How I did it
Use share memory to support counters.
Move check for EnableTranslibWrite to beginning of Set(). This change will simplify the logic for gnmi set fail, and no need to authenticate when EnableTranslibWrite  is false. 

#### How to verify it
Build image and run unit test.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/common_utils/context.go
+++ b/common_utils/context.go
@@ -36,6 +36,16 @@ const requestContextKey contextkey = 0
 // Request Id generator
 var requestCounter uint64
 
+var CountersName = [...]string {
+	"GNMI get",
+	"GNMI get fail",
+	"GNMI set",
+	"GNMI set fail",
+}
+
+var globalCounters [len(CountersName)]uint64
+
+
 // GetContext function returns the RequestContext object for a
 // gRPC request. RequestContext is maintained as a context value of
 // the request. Creates a new RequestContext object is not already
@@ -55,8 +65,25 @@ func GetContext(ctx context.Context) (*RequestContext, context.Context) {
 
 func GetUsername(ctx context.Context, username *string) {
 	rc, _ := GetContext(ctx)
-        if rc != nil {
-            *username = rc.Auth.User
-        }
+	if rc != nil {
+		*username = rc.Auth.User
+	}
+}
+
+func InitCounters() {
+	for i := 0; i < len(CountersName); i++ {
+		globalCounters[i] = 0
+	}
+	SetMemCounters(&globalCounters)
+}
+
+func IncCounter(name string) {
+	for i := 0; i < len(CountersName); i++ {
+		if CountersName[i] == name {
+			atomic.AddUint64(&globalCounters[i], 1)
+			break
+		}
+	}
+	SetMemCounters(&globalCounters)
 }
 
--- a/common_utils/context.go
+++ b/common_utils/context.go
@@ -36,14 +36,31 @@ const requestContextKey contextkey = 0
 // Request Id generator
 var requestCounter uint64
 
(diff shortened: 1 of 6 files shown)


gnmi_server/server.go
```diff
@@ -388,13 +404,11 @@ func (s *Server) Set(ctx context.Context, req *gnmipb.SetRequest) (*gnmipb.SetRe
 		/* Add to Set response results. */
 		results = append(results, &res)
 	}
-	if s.config.EnableTranslibWrite {
```
qiluo-msft, line 391: >EnableTranslibWrite [](http://example.com/codeflow?start=13&length=19)

Why removing this check?

ganglyu: I move the check to line 353, when EnableTranslibWrite is false, no need to authenticate and no need to update result.

common_utils/context.go
```diff
@@ -55,8 +65,25 @@ func GetContext(ctx context.Context) (*RequestContext, context.Context) {
...
+}
+
+func InitCounters() {
+	for i := 0; i < len(CountersName); i++ {
+		globalCounters[i] = 0
+	}
+	SetMemCounters(&globalCounters)
+}
+
+func IncCounter(name string) {
```
tomek-US, line 80: Why use string as an index? If an enum were used then there would be no need for the loop in line 81, right?

gnmi_server/server_test.go
```diff
@@ -107,6 +107,25 @@ func createServer(t *testing.T, port int64) *Server {
 	return s
 }
 
+func createReadServer(t *testing.T, port int64) *Server {
+	certificate, err := testcert.NewCert()
+	if err != nil {
+		t.Errorf("could not load server key pair: %s", err)
```
tomek-US, line 113: This should be `t.Fatalf(...)` as `t.Errorf(...)` does not stop execution of the test and it will continue despite lack of the certificate.

gnmi_server/server_test.go
```diff
@@ -107,6 +107,25 @@ func createServer(t *testing.T, port int64) *Server {
...
+	tlsCfg := &tls.Config{
+		ClientAuth:   tls.RequestClientCert,
+		Certificates: []tls.Certificate{certificate},
+	}
+
+	opts := []grpc.ServerOption{grpc.Creds(credentials.NewTLS(tlsCfg))}
+	cfg := &Config{Port: port, EnableTranslibWrite: false}
+	s, err := NewServer(cfg, opts)
+	if err != nil {
+		t.Errorf("Failed to create gNMI server: %v", err)
```
tomek-US, line 124: Similar comment here. This should be `t.Fatalf(...)` instead of `t.Errorf(...)`

gnmi_server/server_test.go
```diff
@@ -107,6 +107,25 @@ func createServer(t *testing.T, port int64) *Server {
 	return s
 }
 
+func createReadServer(t *testing.T, port int64) *Server {
+	certificate, err := testcert.NewCert()
```
tomek-US, line 111: As this is a helper function it should be marked as such by calling `t.Helper()` at its beginning.

qiluo-msft (approved):
LGTM. Please also check with other reviewers.

ganglyu:
@tomek-US I have updated, would you like to review again?

PR #61, opened 2022-11-15
Fix heartbeat not stopping after client closes

zbud-msft:
#### Why I did it

When client conn closes, returning due to queue error in event_send will not allow event_deinit_sub to occur, which will not stop hearbeat/start cache

#### How I did it

Change return to break

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


renukamanavalan (approved):
Excellent find.

PR #64, opened 2022-11-24
Call evtc_stop after error

zbud-msft:
#### Why I did it

Currently after error, we don't check to see if evtc has stopped via chan, need to make sure if evtc stopped then we call deinit subscriber

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


sonic_data_client/events_client.go
```diff
@@ -332,7 +332,8 @@ func get_events(evtc *EventClient) {
                                 JsonIetfVal: jv,
                             }}
                         if err := send_event(evtc, evtTv, evt.Publish_epoch_ms); err != nil {
-                            break
+                            check_evtc_stopped(evtc)
```
ganglyu, line 335: Why not check the result? Can you explain?

zbud-msft: Because we will return if there is an error regardless whether or not event client has been stopped. In the other call to this function we check if event client has been stopped becuase if it has then we will also break out. With the call to this function, we mainly just want to see if event client has been stopped, deinit subscriber.

ganglyu: If event client is not stopped, when will you deinit subscribe?

zbud-msft: We only deinit subscriber when event client is stopped, as the subscriber will subscribe to keep listening for events for as long as they want to

zbud-msft:
@renukamanavalan Now in main thread StreamRun, it will now keep checking channel only if evtc.stopped == 0, as send event failure inside get_events will also set evtc.stopped = 1, as we want to exit all go routines and to stop looking for close, essentially perform our own close operation

PR #68, opened 2022-12-31
Add dataset for output of "df -h" into non-db-client

isabelmsft:
#### Why I did it
Disk utilization output is currently missing in telemetry.
The disk utilization output by snmp is inaccurate - it reports for the snmp docker, and not the SONiC host machine.

#### How I did it
- Add path for gnmi to GET disk_utilization info
- Create sonic-host-services PR to add host service which reads output of `df -h` https://github.com/sonic-net/sonic-host-services/pull/29
- Connect gnmi GET disk_utilization API to dbus service
- Connect new host service which reads output of `df -h` to dbus service
- From gnmi GET API, call into dbus API that retrieves disk utilization info

#### How to verify it
Use gnmi_cli to read from the disk/usage path in OTHER dataset

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/sonic_service_client/dbus_client.go
+++ b/sonic_service_client/dbus_client.go
@@ -16,6 +16,7 @@ type Service interface {
 	ApplyPatchDb(fileName string) error
 	CreateCheckPoint(cpName string)  error
 	DeleteCheckPoint(cpName string) error
+	GetDiskUtilization() (error, []map[string]string)
 }
 
 type DbusClient struct {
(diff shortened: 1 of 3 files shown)


ganglyu:
Please add unit test for your change.

sonic_service_client/dbus_client.go
```diff
@@ -147,3 +204,19 @@ func (c *DbusClient) DeleteCheckPoint(fileName string) error {
 	err := DbusApi(busName, busPath, intName, 10, fileName)
 	return err
 }
+
+func (c *DbusClient) GetDiskUtilization() (error, []map[string]string) {
```
ganglyu, line 208: I guess DBUS return value is not []map[string]string

PR #69, opened 2023-01-04
Fix gnmi cli hang

zbud-msft:
#### Why I did it

gnmi_cli will hang when trying to write to channel inside cli_config. Use one goroutine to write to term channel and check if received cnt == expected cnt, otherwise when sleep_cnt hits max term due to timeout

#### How to verify it

Nightly tests

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


patches/0001-Fix-gnmi_cli-hang.patch
```diff
@@ -0,0 +1,47 @@
+From be4d7a0f0c902fc5fc7cc4dbc2f8342679b4daf9 Mon Sep 17 00:00:00 2001
+From: zbud-msft <zainbudhwani@microsoft.com>
+Date: Wed, 4 Jan 2023 00:07:04 +0000
+Subject: [PATCH] Fix gnmi_cli hang
+
```
ganglyu, line 5: Please add comment to explain your patch

Makefile
```diff
@@ -53,6 +53,7 @@ $(GO_DEPS): go.mod $(PATCHES) swsscommon_wrap
 	patch -d vendor -p0 < patches/gnmi_path.patch
 	patch -d vendor -p0 < patches/gnmi_xpath.patch
 	git apply patches/0001-Updated-to-filter-and-write-to-file.patch
+	git apply patches/0001-Fix-gnmi_cli-hang.patch
```
renukamanavalan, line 56: This should be "0002-...."

renukamanavalan (approved):

Makefile
zbud-msft: Combined into one patch

patches/0001-Fix-gnmi_cli-hang.patch
zbud-msft: Fixed last patch by removing writing to channel inside cli_config. Combined fixes into existing patch

ganglyu (approved):

PR #80, opened 2023-02-27
Add logs for md5 checksum

zbud-msft:
#### Why I did it

Add additional logs for when there is a mismatching public key and private key such as timestamp and md5 checksum

#### How I did it

Add logs

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [x] 202012
- [ ] 202106
- [ ] 202111
- [x] 202205

### DIFF
--- a/telemetry/telemetry.go
+++ b/telemetry/telemetry.go
@@ -92,8 +92,8 @@ func main() {
 			certificate, err = tls.LoadX509KeyPair(*serverCert, *serverKey)
 			if err != nil {
 				currentTime := time.Now()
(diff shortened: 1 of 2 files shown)


telemetry/telemetry.go
```diff
@@ -90,6 +91,9 @@ func main() {
 			}
 			certificate, err = tls.LoadX509KeyPair(*serverCert, *serverKey)
 			if err != nil {
+				currentTime := time.Now()
```
qiluo-msft, line 94: >Now() [](http://example.com/codeflow?start=24&length=5)

always use UTC?

zbud-msft: Changed to UTC

qiluo-msft (approved):

PR #96, opened 2023-04-21
Implementation of Master Arbitration.

sallylsy:

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


gnmi_server/server.go
```diff
@@ -514,3 +526,56 @@ func (s *Server) Capabilities(ctx context.Context, req *gnmipb.CapabilityRequest
 		GNMIVersion:        "0.7.0",
 		Extension:          exts}, nil
 }
+
+type uint128 struct {
+	High uint64
+	Low  uint64
+}
+
+func (lh *uint128) Compare(rh *uint128) int {
```
ayelrod, line 535: Check if lh or rh are nil at the start of Compare?

gnmi_server/server_test.go
```diff
@@ -291,10 +291,16 @@ func runTestSet(t *testing.T, ctx context.Context, gClient pb.GNMIClient, pathTa
 		req = &pb.SetRequest{
 			Prefix: &prefix,
 			Replace: []*pb.Update{&pb.Update{Path: &pbPath, Val: v}},
+			Extension: []*ext_pb.Extension{
+				{Ext: &ext_pb.Extension_MasterArbitration{}},
```
ayelrod, line 295: If master arbitration is disabled, are these extensions necessary?

gnmi_server/server.go
sallylsy: rh nil check is needed. If lh is nil, the error should be handled when the function get called.

gnmi_server/server_test.go
sallylsy: No. Extension field is optional

ayelrod (approved):

gnmi_server/server_test.go
```diff
@@ -3400,11 +3407,152 @@ func TestParseOrigin(t *testing.T) {
...
+		}
+	})
+	// After this test ElectionID is one.
+	t.Run("MasterArbitrationOnElectionIdZeroThenOne", func(t *testing.T) {
+		req := &pb.SetRequest{
+			Update:    []*pb.Update{update},
+			Extension: []*ext_pb.Extension{maExt0},
+		}
+		_, err = gClient.Set(ctx, req)
+		if _, ok := status.FromError(err); !ok {
```
sachinholla, line 3481: This check does not catch the MA failures.. Please fix here and everywhere else

gnmi_server/server_test.go
```diff
@@ -3400,11 +3407,152 @@ func TestParseOrigin(t *testing.T) {
...
+			t.Fatal("Got a non-grpc error from grpc call")
+		}
+		req = &pb.SetRequest{
+			Update:    []*pb.Update{update},
+			Extension: []*ext_pb.Extension{maExt1},
+		}
+		_, err = gClient.Set(ctx, req)
+		if _, ok := status.FromError(err); !ok {
+			t.Fatal("Got a non-grpc error from grpc call")
+		}
```
sachinholla, line 3491: Also check the updated `s.masterEID` value for all success cases

gnmi_server/server.go
```diff
@@ -515,3 +527,63 @@ func (s *Server) Capabilities(ctx context.Context, req *gnmipb.CapabilityRequest
...
+		ma := e.GetMasterArbitration()
+		if ma == nil {
+			continue
+		}
+		// The Master Arbitration descriptor has been found.
+		// If MA is enabled, the reqEID needs to be set to 0
+		if ma.ElectionId != nil {
+			reqEID = uint128{High: ma.ElectionId.High, Low: ma.ElectionId.Low}
+		}
+		// Use the election ID that is in the last extension, so, no 'break' here.
```
sachinholla, line 572: Is it better to iterate backwards to pick the last value?

gnmi_server/server.go
```diff
@@ -515,3 +527,63 @@ func (s *Server) Capabilities(ctx context.Context, req *gnmipb.CapabilityRequest
...
+	}
+	switch masterEID.Compare(&reqEID) {
```
sachinholla, line 574: Since `reqEID` was inited with {0,0} value, this check fails when the current request does not have any MA extension but previous request had set one. But the [spec](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-master-arbitration.md#32-setrequest-handling) wants servers to access such requests:

> If there is no arbitration extension associated with the Set request, the Set request should be accepted and processed. Therefore, client that expects to ever be part of a replica-group MUST always set this extension, and the role MUST be populated.

gnmi_server/server.go
```diff
@@ -515,3 +527,63 @@ func (s *Server) Capabilities(ctx context.Context, req *gnmipb.CapabilityRequest
...
+	case 1: // This Election ID is smaller than the known Master Election ID.
+		log.V(1).Infoln("Election ID is smaller than the current master. Rejected.")
+		return false
+	case -1: // New Master Election ID received!
+		log.V(0).Infof("New master has been elected with %v\n", reqEID)
+		*masterEID = reqEID
```
sachinholla, line 580: This can create a race when two set requests arrive at the same time. Better to use atomic operations

gnmi_server/server_test.go
```diff
@@ -3400,11 +3407,152 @@ func TestParseOrigin(t *testing.T) {
...
+		if ret.Code() != codes.PermissionDenied {
+			t.Fatalf("Expected PermissionDenied. Got %v", ret.Code())
+		}
+		req = &pb.SetRequest{
+			Update:    []*pb.Update{update},
+			Extension: []*ext_pb.Extension{},
+		}
+		_, err = gClient.Set(ctx, req)
+		if err == nil {
+			t.Fatal("Expected a PermissionDenied error")
```
sachinholla, line 3539: Requests without MA extension should always pass

telemetry/telemetry.go
```diff
@@ -33,6 +33,7 @@ var (
 	gnmi_translib_write = flag.Bool("gnmi_translib_write", gnmi.ENABLE_TRANSLIB_WRITE, "Enable gNMI translib write for management framework")
 	gnmi_native_write   = flag.Bool("gnmi_native_write", gnmi.ENABLE_NATIVE_WRITE, "Enable gNMI native write")
 	threshold         = flag.Int("threshold", 100, "max number of client connections")
+	withMasterArbitration = flag.Bool("with-master-arbitration", false, "Enables master arbitration policy.")
```
sachinholla, line 36: How an admin can enable `--with-master-arbitration` flag? I did not find any PR to update the telemetry service startup script

sachinholla (changes requested):

gnmi_server/server.go
```diff
@@ -515,3 +527,63 @@ func (s *Server) Capabilities(ctx context.Context, req *gnmipb.CapabilityRequest
...
+	// Read the election_id.
+	reqEID := uint128{High: 0, Low: 0}
+	// It can be one of many extensions, so iterate through them to find it.
+	for _, e := range req.GetExtension() {
+		ma := e.GetMasterArbitration()
+		if ma == nil {
+			continue
+		}
+		// The Master Arbitration descriptor has been found.
+		// If MA is enabled, the reqEID needs to be set to 0
```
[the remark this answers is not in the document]
sallylsy: I can re-word it as following:

// If MA is enabled, the reqEID is initialized with 0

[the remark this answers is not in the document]
sallylsy: I will re-word it to:

If MA is enabled, reqEID will be inited with {0,0} value

telemetry/telemetry.go
sachinholla: *My original comment got truncated for some reason.. Following is the continuation of it:*

What will happen if client sends MA extension and server is not started with `--with-master-arbitration` flag? Current implementation ignores the MA extension and accepts all set requests. Clients cannot discover that  that MA is off and we may enter an incorrect state should a management network partition happen.

I feel we do not need this flag. The spec already gives full control to the clients to use MA or not. Set requests will be accepted if client did not send a MA extension. So, legacy clients will continue to work as before. This flag is redundant.

gnmi_server/server.go
sallylsy: Done

gnmi_server/server.go
```diff
@@ -515,3 +527,63 @@ func (s *Server) Capabilities(ctx context.Context, req *gnmipb.CapabilityRequest
...
+		if ma.ElectionId != nil {
+			reqEID = uint128{High: ma.ElectionId.High, Low: ma.ElectionId.Low}
+		}
```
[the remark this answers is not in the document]
sallylsy: Done

gnmi_server/server.go
```diff
@@ -515,3 +527,63 @@ func (s *Server) Capabilities(ctx context.Context, req *gnmipb.CapabilityRequest
...
+}
+
+// ReqFromMasterEnabledMA returns true if the request is sent by the master
+// controller.
+func ReqFromMasterEnabledMA(req *gnmipb.SetRequest, masterEID *uint128) bool {
+	// Read the election_id.
+	reqEID := uint128{High: 0, Low: 0}
+	// It can be one of many extensions, so iterate through them to find it.
+	for _, e := range req.GetExtension() {
+		ma := e.GetMasterArbitration()
```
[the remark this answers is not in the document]
sallylsy: Done. Please see L572.

gnmi_server/server.go
sallylsy: I add mutex here to prevent racing issue.

gnmi_server/server_test.go
sallylsy: Done

gnmi_server/server.go
sachinholla: This local mutex will not guard against parallel set calls. It should be made a global variable.
Also, i feel the `masterEID.Compare()` call also should be guarded.

gnmi_server/server_test.go
```diff
@@ -3400,11 +3407,181 @@ func TestParseOrigin(t *testing.T) {
...
+		ret, ok := status.FromError(err)
+		if !ok {
+			t.Fatal("Got a non-grpc error from grpc call")
+		}
+		if ret.Code() != codes.PermissionDenied {
+			t.Fatalf("Expected PermissionDenied. Got %v", ret.Code())
+		}
+		reqEid0 := maExt0.GetMasterArbitration().GetElectionId()
+		expectedEID0 := uint128{High: reqEid0.GetHigh(), Low: reqEid0.GetLow()}
+		if s.masterEID.Compare(&expectedEID0) != 0 {
```
sachinholla, line 3561: s.masterEID will be {1, 0} here

gnmi_server/server_test.go
```diff
@@ -3400,11 +3407,181 @@ func TestParseOrigin(t *testing.T) {
...
+		req = &pb.SetRequest{
+			Update:    []*pb.Update{update},
+			Extension: []*ext_pb.Extension{},
+		}
+		_, err = gClient.Set(ctx, req)
+		if err != nil {
+			t.Fatal("Expected a successful set call.")
+		}
+		reqEid0 := maExt0.GetMasterArbitration().GetElectionId()
+		expectedEID0 := uint128{High: reqEid0.GetHigh(), Low: reqEid0.GetLow()}
```
sachinholla, line 3573: s.masterEID will be {1, 0} here

gnmi_server/server.go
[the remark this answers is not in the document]
sallylsy: Corrected.

gnmi_server/server.go
```diff
@@ -515,3 +527,63 @@ func (s *Server) Capabilities(ctx context.Context, req *gnmipb.CapabilityRequest
...
+		// The Master Arbitration descriptor has been found.
+		// If MA is enabled, the reqEID needs to be set to 0
+		if ma.ElectionId != nil {
+			reqEID = uint128{High: ma.ElectionId.High, Low: ma.ElectionId.Low}
+		}
+		// Use the election ID that is in the last extension, so, no 'break' here.
+	}
+	switch masterEID.Compare(&reqEID) {
+	case 1: // This Election ID is smaller than the known Master Election ID.
+		log.V(1).Infoln("Election ID is smaller than the current master. Rejected.")
```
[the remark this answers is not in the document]
sallylsy: Done.

gnmi_server/server.go
```diff
@@ -515,3 +527,71 @@ func (s *Server) Capabilities(ctx context.Context, req *gnmipb.CapabilityRequest
...
+		if ma == nil {
+			continue
+		}
+		// The Master Arbitration descriptor has been found.
+		if ma.ElectionId == nil {
+			return status.Error(codes.InvalidArgument, "No election id provided by master.")
+		}
+
+		if ma.Role != nil {
+			return status.Error(codes.Unimplemented, "Role is not implemented.")
```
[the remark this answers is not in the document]
sallylsy: Corrected.

gnmi_server/server.go
sallylsy: Added mutex on EID.

gnmi_server/server.go
```diff
@@ -515,3 +527,76 @@ func (s *Server) Capabilities(ctx context.Context, req *gnmipb.CapabilityRequest
...
+	// It can be one of many extensions, so iterate through them to find it.
+	for _, e := range req.GetExtension() {
+		ma := e.GetMasterArbitration()
+		if ma == nil {
+			continue
+		}
+		// The Master Arbitration descriptor has been found.
+		if ma.ElectionId == nil {
+			log.V(1).Infoln("No election id provided by master.")
+			continue
```
sachinholla, line 573: Spec expects InvalidArgument error for this case, but I guess the code results in a PermissionDenied error. Better to change signature of ReqFromMasterEnabledMA function to return an error.

```go
func ReqFromMasterEnabledMA(req *gnmipb.SetRequest, masterEID *uint128) error {
    ....
    if ma.ElectionId == nil {
        return status.Errorf(codes.InvalidArgument, "MA: ElectionId missing")
    }
    if ma.Role != nil {
        return status.Errorf(codes.Unimplemented, "MA: Role is not implemented")
    }
    ....
    ...
    if request EID < master EID {
        return status.Errorf(codes.PermissionDenied, "Not a master with election ID %v", masterEID)
    }
}

func (s *Server) Set(ctx context.Context, req *gnmipb.SetRequest) (*gnmipb.SetResponse, error) {
    if err := s.ReqFromMaster(req, &s.masterEID); err != nil {
        return nil, err
    }
    ....
```

gnmi_server/server.go
sachinholla: I think this will not work. It will overwrite the mutex in masterEID as well and the Unlock() call in next line will fail. Have you verified this?

Also, this code intends to lock/unlock the mutex separately for EID compare and update operations. Ideally both should be one atomic operation. Otherwise there is a chance (in theory) of EID getting overwritten by another goroutine after the compare was done but before the value is updated. Better to synchronize the entire ReqFromMasterEnabledMA logic, something like:

```go

var maMu sync.Mutex

func ReqFromMasterEnabledMA(req *gnmipb.SetRequest, masterEID *uint128) error {
    /*** find reqEID ***/
    maMu.Lock()
    defer maMu.Unlock()
    switch masterEID.Compare(reqEID) {
    case -1:
        *masterEID = reqEID
    ....
    ...
```

sallylsy: Done.

gnmi_server/server_test.go
```diff
@@ -3400,11 +3407,152 @@ func TestParseOrigin(t *testing.T) {
...
+	}
+
+	// By default ElectionID starts from 0 so this test does not change it.
+	t.Run("MasterArbitrationOnElectionIdZero", func(t *testing.T) {
+		req := &pb.SetRequest{
+			Update:    []*pb.Update{update},
+			Extension: []*ext_pb.Extension{maExt0},
+		}
+		_, err = gClient.Set(ctx, req)
+		if _, ok := status.FromError(err); !ok {
```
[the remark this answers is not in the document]
sallylsy: Added `err != nil` check.

gnmi_server/server.go
sallylsy: Thanks. I changed the check here. If the MA extension does not exist, EID will not be set.

gnmi_server/server_test.go
sallylsy: Done.

gnmi_server/server.go
sallylsy: Thanks. func has been modified.

gnmi_server/server.go
sallylsy: Since we have checked is `ma == nil` previously, the last available MA extension should have its EID updated here.

gnmi_server/server.go
```diff
@@ -377,6 +386,28 @@ func (s *Server) Get(ctx context.Context, req *gnmipb.GetRequest) (*gnmipb.GetRe
 }
 
 func (s *Server) Set(ctx context.Context, req *gnmipb.SetRequest) (*gnmipb.SetResponse, error) {
+	e := s.ReqFromMaster(req, &s.masterEID)
+	if e != nil {
+		ret, ok := status.FromError(e)
```
sachinholla, line 391: Why do we these error type checks? Error message can be formatted correctly at the source instead of unboxing and re-boxing here.

gnmi_server/server.go
```diff
@@ -515,3 +546,79 @@ func (s *Server) Capabilities(ctx context.Context, req *gnmipb.CapabilityRequest
...
+		log.V(0).Infof("New master has been elected with %v\n", reqEID)
+		*masterEID = reqEID
+	}
+	return nil
+}
+
+// ReqFromMasterDisabledMA always returns true. It is used when Master Arbitration
+// is disabled.
+func ReqFromMasterDisabledMA(req *gnmipb.SetRequest, masterEID *uint128) error {
+	return status.Errorf(codes.Unavailable, "MA: Master Arbitration is disabled.")
```
sachinholla, line 623: `return nil` would be better

gnmi_server/server_test.go
```diff
@@ -3610,11 +3610,202 @@ func TestParseOrigin(t *testing.T) {
...
+			Update: []*pb.Update{
+				newPbUpdate("interface[name=Ethernet0]/config/mtu", `{"mtu": 9104}`),
+			},
+			Extension: []*ext_pb.Extension{maExt0},
+		}
+		_, err = gClient.Set(ctx, req)
+		if err != nil {
+			t.Fatal("Did not expected an error: " + err.Error())
+		}
+		if _, ok := status.FromError(err); !ok {
```
sachinholla, line 3675: This condition can be removed. It will always be false due to the nil check above.

sallylsy: I think we should be good. 
`If err is nil, a Status is returned with codes.OK and no message.`
Ref: https://pkg.go.dev/google.golang.org/grpc/status#FromError

gnmi_server/server.go
sallylsy: Done

gnmi_server/server.go
sallylsy: Done

gnmi_server/server_test.go
```diff
@@ -3610,11 +3610,202 @@ func TestParseOrigin(t *testing.T) {
...
+			t.Fatal("Got a non-grpc error from grpc call")
+		}
+		reqEid10 := maExt1H0L.GetMasterArbitration().GetElectionId()
+		expectedEID10 := uint128{High: reqEid10.GetHigh(), Low: reqEid10.GetLow()}
+		if s.masterEID.Compare(&expectedEID10) != 0 {
+			t.Fatalf("Master EID update failed. Want %v, got %v", expectedEID10, s.masterEID)
+		}
+	})
+	// As the ElectionID is one, a request with ElectionID==0 will fail.
+	// Also a request without Election ID will fail.
```
sachinholla, line 3762: typo -- request without EID will **succeed**

sachinholla (approved):

tomek-US (approved):
LGTM

PR #99, opened 2023-04-26
Add keep alive policies to close idle client connections

zbud-msft:
#### Why I did it

In integration test, we have a gnmi client running in PTF docker that is creating 5000 consecutive subscribe requests, sample mode, in which after receiving a single update, the client is no longer making rPC requests. The connection is still alive on the server-side and you can see below the top 3 goroutines balloon up all the way to ~1784 alive connections, which causes a high memory utilization alert inside the telemetry container. 

Inside sonic-gnmi code, I tested with simple http server that profiles the current process using pprof. Pprof will capture state of the process every few seconds (heap size, alloc, goroutine, etc). From the beginning of the process, when there were fewer client connections, we could see the top 3 connections at around ~500, but near the end see that it goes to 1784. Proof that these connections are still being held alive even after client is finished gathering data. 

The reason why we don't see this issue as much in production is that the default keep alive for the server is 2 hours, meaning idle connections will eventually start to close at around 2 hours, releasing this allocated memory.  

#### How I did it

[Issue #1269 · grpc/grpc-go (github.com)](https://github.com/grpc/grpc-go/issues/1269)
[Issue #2086 · grpc/grpc-go (github.com)](https://github.com/grpc/grpc-go/issues/2086)

From researching similar issues in github.com/grpc/grpc-go, we find that we can close idle connections after some specific time by setting MaxConnectionIdle

Refer: https://github.com/grpc/grpc-go/blob/master/keepalive/keepalive.go#L50

#### How to verify it

End to end integration test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/gnmi_server/server.go
+++ b/gnmi_server/server.go
@@ -21,32 +21,14 @@ import (
 	"google.golang.org/grpc/peer"
 	"google.golang.org/grpc/reflection"
 	"google.golang.org/grpc/status"
-	"google.golang.org/grpc/keepalive"
 	"net"
 	"strings"
 	"sync"
-	"time"
 )
 
 var (
 	supportedEncodings = []gnmipb.Encoding{gnmipb.Encoding_JSON, gnmipb.Encoding_JSON_IETF}
 )
-
-var (
-	keep_alive_enforcement = keepalive.EnforcementPolicy{
-		MinTime:            5 * time.Second,
-		PermitWithoutStream: true,
-	}
-)
-
-var (
-	keep_alive_params = keepalive.ServerParameters{
-		MaxConnectionIdle: 5 * time.Second,
-		MaxConnectionAge:  120 * time.Second,
-		MaxConnectionAgeGrace: 30 * time.Second,
-	}
-)
-
 
 // Server manages a single gNMI Server implementation. Each client that connects
 // via Subscribe or Get will receive a stream of updates based on the requested
@@ -70,6 +52,7 @@ type Config struct {
 	UserAuth AuthTypes
 	EnableTranslibWrite bool
 	EnableNativeWrite bool
+	IdleConnDuration int
 }
 
 var AuthLock sync.Mutex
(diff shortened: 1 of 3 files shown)


zbud-msft:
MicrosoftTeams-image (6)

ganglyu:
How to verify? Is it possible to add unit test?

ganglyu:
Can you add end2end test, and then we can review 2 PRs at once.

gnmi_server/server.go
```diff
@@ -21,15 +21,33 @@ import (
...
+		MinTime:            5 * time.Second,
+		PermitWithoutStream: true,
+	}
+)
+
+var (
+	keep_alive_params = keepalive.ServerParameters{
+		MaxConnectionIdle: 5 * time.Second,
+		MaxConnectionAge:  120 * time.Second,
+		MaxConnectionAgeGrace: 30 * time.Second,
```
qiluo-msft, line 46: Could you explain the meaning of these parameters you are choosing? And what is the default values for current implementation so we can understand you decrease(increase) some values.

Is it better to move hard code constants into a config file? so in runtime we still have opportunity to tune them.

gnmi_server/server.go
```diff
@@ -21,15 +21,33 @@ import (
...
 
+var (
+	keep_alive_enforcement = keepalive.EnforcementPolicy{
+		MinTime:            5 * time.Second,
+		PermitWithoutStream: true,
+	}
+)
+
+var (
+	keep_alive_params = keepalive.ServerParameters{
```
qiluo-msft, line 43: >ServerParameters [](http://example.com/codeflow?start=31&length=16)

Does any parameter impact on-demand subscription? For example if subscribing a very stable counter, and there is no counter change for days, will it be disconnected?

gnmi_server/server.go
zbud-msft: MaxConnectionIdle as defined in grpc library, "MaxConnectionIdle is a duration for the amount of time after which an idle connection would be closed by sending a GoAway. Idleness duration is defined since the most recent time the number of outstanding RPCs became zero or the connection establishment."

Default value is infinity.

I will do the same with idle_conn_duration, have telemetry.sh define value if not in config_db.

gnmi_server/server.go
zbud-msft: No, it will not be disconnected. This change only affects closing idle connections, which means ON_CHANGE connections won't be disconnected even if there are no updates. Tested ON_CHANGE with "STATE_DB/DEVICE_METADATA" where no updates were happening. Connection did not close.

qiluo-msft (approved):

PR #108, opened 2023-05-09
Add SaveOnSet

rlucus:
#### Why I did it
Enable save-on-set for gnmi. Creates a configDB backup to disk after eat gnmi.Set if enabled.

#### How I did it
Add a function pointer that allows the entire feature to be enabled by command line argument. If enabled, a call to sonic-services-client sends a message on the DBUS to the sonic host service modules.

HLD here
https://github.com/sonic-net/SONiC/issues/1239

sonic-buildimage PR here
https://github.com/sonic-net/sonic-buildimage/pull/18481


gnmi_server/server.go
```diff
@@ -376,6 +382,18 @@ func (s *Server) Get(ctx context.Context, req *gnmipb.GetRequest) (*gnmipb.GetRe
 	return &gnmipb.GetResponse{Notification: notifications}, nil
 }
 
+// SaveOnSetEnabled saves configuration to a file
+func SaveOnSetEnabled() {
+	if err := transformer.SaveStartupConfig(); err != nil {
```
tomek-US, line 387:
```
gnmi_server/server.go:387:12: undefined: transformer.SaveStartupConfig
```

rlucus: Its dependent on https://github.com/sonic-net/sonic-mgmt-common/pull/83

gnmi_server/server.go
```diff
@@ -376,6 +382,18 @@ func (s *Server) Get(ctx context.Context, req *gnmipb.GetRequest) (*gnmipb.GetRe
 	return &gnmipb.GetResponse{Notification: notifications}, nil
 }
 
+// SaveOnSetEnabled saves configuration to a file
+func SaveOnSetEnabled() {
```
anand-kumar-subramanian, line 386: Do you need this function to be exposed to other packages? Shouldn't this be a local only function?

anand-kumar-subramanian (changes requested):

gnmi_server/server.go
anand-kumar-subramanian, line 387: Should not invoke transformer functions directly from here. Always go through transl and translib.

gnmi_server/server.go
```diff
@@ -5,16 +5,17 @@ import (
 	"errors"
 	"fmt"
 	"github.com/Azure/sonic-mgmt-common/translib"
-	"github.com/sonic-net/sonic-gnmi/common_utils"
-	spb "github.com/sonic-net/sonic-gnmi/proto"
-	spb_gnoi "github.com/sonic-net/sonic-gnmi/proto/gnoi"
-	spb_jwt_gnoi "github.com/sonic-net/sonic-gnmi/proto/gnoi/jwt"
-	sdc "github.com/sonic-net/sonic-gnmi/sonic_data_client"
+	"github.com/Azure/sonic-mgmt-common/translib/transformer"
```
anand-kumar-subramanian, line 8: Should not include transformer package here. Only interface between the gNMI server management common packages should be through translib

gnmi_server/server.go
rlucus: If I understand correctly, this would require a much deeper change to support. I've currently modified according to anand's and sachin's request to place SaveStartupConfig in translib instead of transformer.

gnmi_server/server.go
```diff
@@ -376,6 +382,18 @@ func (s *Server) Get(ctx context.Context, req *gnmipb.GetRequest) (*gnmipb.GetRe
 	return &gnmipb.GetResponse{Notification: notifications}, nil
 }
 
+// SaveOnSetEnabled saves configuration to a file
+func SaveOnSetEnabled() {
+	if err := transformer.SaveStartupConfig(); err != nil {
+		log.Errorf("Saving startup config failed: %v", err)
+	} else {
+		log.Errorf("Success! Startup config has been saved!")
```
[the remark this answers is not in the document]
rlucus: Done

gnmi_server/server.go
rlucus: Done

gnmi_server/server.go
rlucus: Done

gnmi_server/server.go
rlucus: Done

gnmi_server/server.go
rlucus: Implemented translib rpc for copy.

anand-kumar-subramanian (dismissed):

anand-kumar-subramanian (approved):

rlucus:
Sorry wrong link to HLD. it is here https://github.com/sonic-net/SONiC/pull/1297

gnmi_server/server.go
```diff
@@ -503,6 +530,7 @@ func (s *Server) Set(ctx context.Context, req *gnmipb.SetRequest) (*gnmipb.SetRe
 		common_utils.IncCounter(common_utils.GNMI_SET_FAIL)
 	}
 
+	s.SaveStartupConfig()
```
sachinholla, line 533: No locks in both Set() and SaveOnSetEnabled(). Is it safe to allow parallel db writes and config saves?

PR #110, opened 2023-05-10
Return Grpc.Internal error once there's socker read timeout issue occurs in streaming subscribe under routing flap case

FengPan-Frank:
Currently streaming telemetry subscribe will hit some error under routing flap case if it subscribes APPL_DB ROUTE_TABLE, this change will return explicit error code to client without any handling.

pubsub notification to avoid inconsistency introduced by other concurrent cmd

#### Why I did it
When there's DB table flapping happens, like APPL_DB ROUTE_Table, current gnmi server has some performance issue which will result in below error
"read unix @->/var/run/redis/redis.sock: i/o timeout"

We should need to fix this systemically. Now the fix is to return explicit error code as Internal error to client, and later we will do some optimization, and add some retry strategy into the redis function call.

#### How I did it
Return Grpc.Internal error to the client

#### How to verify it
verified via gnmi_cli, meanwhile some tests which introduces bgp flapping.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/gnmi_server/client_subscribe.go
+++ b/gnmi_server/client_subscribe.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"net"
 	"sync"
+	"strings"
 
 	"github.com/Workiva/go-datastructures/queue"
 	log "github.com/golang/glog"
@@ -207,6 +208,10 @@ func (c *Client) Run(stream gnmipb.GNMI_SubscribeServer) (err error) {
 	c.Close()
 	// Wait until all child go routines exited
 	c.w.Wait()
+	if strings.Contains(err.Error(), "i/o timeout") {
+		return grpc.Errorf(codes.Internal, "%s", err)
+	}
+
 	return grpc.Errorf(codes.InvalidArgument, "%s", err)
 }
 
--- a/sonic_data_client/db_client.go
+++ b/sonic_data_client/db_client.go
@@ -80,6 +80,7 @@ var IntervalTicker = func(interval time.Duration) <-chan time.Time {
 }
 
 var NeedMock bool = false
+var MockFail int = 0
 var intervalTickerMutex sync.Mutex
 
 // Define a new function to set the IntervalTicker variable
@@ -742,6 +743,12 @@ func tableData2Msi(tblPath *tablePath, useKey bool, op *string, msi *map[string]
 		makeJSON_redis(msi, &tblPath.jsonTableKey, op, fv)
 		log.V(6).Infof("Added json key %v fv %v ", tblPath.jsonTableKey, fv)
 		return nil
+	}
+
+	if MockFail == 1 {
+		MockFail++
+		fmt.Println("Mock sleep for redis timeout")
+		time.Sleep(30 * time.Second)
 	}
 
 	for idx, dbkey := range dbkeys {
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,0 +1,22 @@
+name: Semgrep
+
+on:
+  pull_request: {}
+  push:
+    branches:
+      - master
+      - '201[7-9][0-1][0-9]'
+      - '202[0-9][0-1][0-9]'
+
+jobs:
+  semgrep:
+    if: github.repository_owner == 'sonic-net'
+    name: Semgrep
+    runs-on: ubuntu-latest
+    container:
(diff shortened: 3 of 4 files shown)


sonic_data_client/db_client.go
```diff
@@ -670,6 +670,8 @@ func emitJSON(v *map[string]interface{}) ([]byte, error) {
 func tableData2Msi(tblPath *tablePath, useKey bool, op *string, msi *map[string]interface{}) error {
 	redisDb := Target2RedisDb[tblPath.dbNamespace][tblPath.dbName]
 
+	// Start a Redis transaction to avoid consistency introduced by concurrent cmd occured.
```
qiluo-msft, line 673: >transaction [](http://example.com/codeflow?start=18&length=11)

The transaction is overkill since it blocked other redis reading/writing for a long time. Suggest only handle exception on HGet.

sallylsy (approved):

sonic_data_client/db_client.go
FengPan-Frank: Changed into ignore instead of blocking the whole serial of operations.

tomek-US:
Please rebase and add test coverage to unblock the merge.

sonic_data_client/db_client.go
```diff
@@ -710,8 +710,8 @@ func tableData2Msi(tblPath *tablePath, useKey bool, op *string, msi *map[string]
 	for idx, dbkey := range dbkeys {
 		fv, err = redisDb.HGetAll(dbkey).Result()
 		if err != nil {
-			log.V(2).Infof("redis HGetAll failed for  %v, dbkey %s", tblPath, dbkey)
-			return err
+			log.V(2).Infof("redis HGetAll failed for  %v, dbkey %s, ignore it since atomic is not guaranteed", tblPath, dbkey)
+			continue
```
qiluo-msft, line 714: >continue [](http://example.com/codeflow?start=3&length=8)

Need unit test

FengPan-Frank:
> Please rebase and add test coverage to unblock the merge.

Done.

sonic_data_client/db_client.go
FengPan-Frank: Done

sonic_data_client/db_client.go
```diff
@@ -744,6 +745,12 @@ func tableData2Msi(tblPath *tablePath, useKey bool, op *string, msi *map[string]
 		return nil
 	}
 
+	if MockFail == 1 {
+		MockFail++
+		fmt.Printf("mock sleep for redis timeout\n")
+		time.Sleep(30 * time.Second)
+	}
```
ganglyu, line 752: Can you use gomonkey to replace MockFail?

gnmi_server/client_subscribe.go
```diff
@@ -207,6 +208,10 @@ func (c *Client) Run(stream gnmipb.GNMI_SubscribeServer) (err error) {
 	c.Close()
 	// Wait until all child go routines exited
 	c.w.Wait()
+	if strings.Contains(err.Error(), "i/o timeout") {
```
qiluo-msft, line 211: >Contains [](http://example.com/codeflow?start=12&length=8)

Match error message string will be fragile? Which library assign this error message? Could you use error code?

FengPan-Frank: Right, but seems this will require NOT small changes, like now we're using priorityQueue in all db_client which only stores error messages, refer https://github.com/sonic-net/sonic-gnmi/blob/01fe667e7d44cdfb2d046064c18e4d14ac6c0b4d/sonic_data_client/db_client.go#L833, do you suggest we polish this part now or later together with the project change?

zbud-msft:
Can you add steps to reproduce issue in description?

sonic_data_client/db_client.go
```diff
@@ -744,6 +745,12 @@ func tableData2Msi(tblPath *tablePath, useKey bool, op *string, msi *map[string]
 		return nil
 	}
 
+	if MockFail == 1 {
```
qiluo-msft, line 748: >MockFail [](http://example.com/codeflow?start=4&length=8)

Should it only be used in unit test?

hdwhdw:
Closing this PR as it has been inactive for two years. Feel free to reopen if needed.

PR #113, opened 2023-06-06
prevent potential panic: return immediately if there exists error

maipbui:
#### Why I did it
Variable `v` is likely modified and later used on error. In some cases this could result in panics due to a nil dereference
#### How I did it
return immediately if there exists error

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


ganglyu:
Please add unit test to cover this change.

sonic_data_client/non_db_client.go
```diff
@@ -566,6 +566,7 @@ func (c *NonDbClient) Get(w *sync.WaitGroup) ([]*spb.Value, error) {
 		v, err := getter()
 		if err != nil {
 			log.V(3).Infof("PollRun getter error %v for %v", err, v)
+			return nil, err
```
ayelrod, line 569: Would it make more sense here to `continue` instead of `return`?

qiluo-msft: I think caller could determine whether retry based on return value of this function.

qiluo-msft (dismissed):

sonic_data_client/client_test.go
```diff
@@ -322,3 +324,28 @@ func TestParseTarget(t *testing.T) {
...
+	}
+
+	// Create a NonDbClient with the mocked dataGetFunc
+	client := &NonDbClient{
+		path2Getter: path2Getter,
+	}
+
+	var w *sync.WaitGroup
+	_, err := client.Get(w)
+	if err == nil {
```
ganglyu, line 348: You can compare the error here.

ganglyu (approved):

PR #128, opened 2023-06-20
[202305] Checkout correct branch from sonic-mgmt-common and sonic-swss-common during pipeline build

sachinholla:
#### Why I did it

Pipeline is always fetching latest of sonic-mgmt-common and sonic-swss-common repos for its builds. They must be use source code from 202305 branch

#### How I did it

Checkout latest from `$(System.PullRequest.TargetBranch)` during PR builds and from `$(Build.SourceBranchName)` during manual builds.

Known issue: other pipeline artefacts are still being downloaded from master branch. Not a blocker as of now.

#### How to verify it

Pipeline build

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

PR #144, opened 2023-08-16
Azp: install sonic yangs during pipline build

sachinholla:
#### Why I did it

Translib infra (sonic-mgmt-common) has an option to import cvl sonic yangs from sonic-yang-models.whl. This option is not enabled right now, but will be in near future. Hence sonic-yang-models.whl needs to be installed on the vm.
Also backport the pipeline build fixes done in 202305 branch.

#### How I did it

* Download & install sonic_yang_models.whl before telemetry build. Translib will start using these sonic yangs
* Download source code and artifacts for the build branch only (https://github.com/sonic-net/sonic-gnmi/pull/128 backport)
* Use pipeline 142 for downloading sonic-buildimage.vs artifacts, which runs on all branches

#### How to verify it

PR build

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


tomek-US (dismissed):
LGTM

tomek-US:
Please rebase. It cannot be merged as it is now.

tomek-US:
@sachinholla The PR cannot be merged as it is reported out-of-date.

azure-pipelines.yml
```diff
@@ -101,6 +101,10 @@ stages:
         sudo dpkg -i ../target/debs/bullseye/libyang*1.0.73*.deb
       displayName: "Install dependency"
 
+    - script: |
```
kwangsuk, line 104: To install coherent packages related to YANG, it would be cleaner if you replace "-script" with the comment "# YANG MODELS" to run after libyang package install. If you want to keep the separate script, please add the same comment.

sachinholla: I prefer keeping it as a separate step.. Will add a comment.. Step's displayName attribute anyway indicates "Install sonic yangs"

kwangsuk (approved):

anand-kumar-subramanian (approved):

PR #174, opened 2023-12-09
Support multiple database for DPU

ganglyu:
#### Why I did it
https://github.com/sonic-net/sonic-buildimage/pull/17161
We have updated database configuration to support multiple databases for smartswitch.
And then we need to update sonic-gnmi to support new database configuration and new GNMI path for smartswitch

#### How I did it
* Support new database configuration

```
{
    "INCLUDES" : [
        {
            "include" : "../../redis/sonic-db/database_config.json"
        },
        {
            "database_name" : "dpu0",
            "include" : "../../redisdpu0/sonic-db/database_config.json"
        }
    ],
    "VERSION" : "1.0"
}
```

Use a single layer instance for each device to support multiple asic and multiple dpu.
For example, for multiple asic, the instances can be named localhost, asic0, asic1, and so on. For multiple dpu, the instances can be named localhost, dpu0, dpu1, and so on. For multiple asic npu and multiple dpu, the instances can be named localhost, asic0, asic1, ..., dpu0, dpu1, and so on.

```
        "DPU_APPL_DB" : {
            "id" : 15,
            "separator": ":",
            "instance" : "redis",
            "format": "proto"
        },
```
We are trying to modify DPU_APPL_DB with GNMI, but the database is not accessible by swsscommon. This is because DPU_APPL_DB is a new database defined in /var/run/redisdpu0/sonic-db/database_config.json. As a workaround, we use GNMI to modify APPL_DB instead, but we need to update swsscommon to support the new database configuration.

* Update GNMI path to support multiple database

I add a new element in GNMI path to support database instance.
```
/APPL_DB/dpu0/DASH_VNET_TABLE/vnet1/
```
The second element is used for target database instance, and it can be localhost, asic0, asic1, ..., dpu0, dpu1, and so on.
If SONiC device is not multi-asic or multi-dpu, this second element should be localhost.

This design is not backward compatible.

#### How to verify it
Run unit test for sonic-gnmi.
Run end to end test for dash and gnmi.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/gnmi_server/server_test.go
+++ b/gnmi_server/server_test.go
@@ -3861,6 +3861,7 @@ print('%s')
 	s.s.Stop()
 }
 
+// Test DPU configuration with multiple databases
 func TestGNMINativeDPU(t *testing.T) {
 	sdcfg.Init()
 	err := test_utils.SetupMultiDPU()
@@ -3893,7 +3894,6 @@ func TestGNMINativeDPU(t *testing.T) {
 	} else {
 		fmt.Println(string(result))
 	}
-
 }
 
 func TestServerPort(t *testing.T) {
--- a/gnmi_server/server_test.go
+++ b/gnmi_server/server_test.go
@@ -3835,7 +3835,9 @@ print('%s')
 	path = filepath.Dir(path)
 
 	var cmd *exec.Cmd
-	cmd = exec.Command("bash", "-c", "cd "+path+" && "+"pytest -m 'not dpu'")
+	// This test is used for single database configuration
+	// Run tests not marked with multidb
+	cmd = exec.Command("bash", "-c", "cd "+path+" && "+"pytest -m 'not multidb'")
 	if result, err := cmd.Output(); err != nil {
 		fmt.Println(string(result))
 		t.Errorf("Fail to execute pytest: %v", err)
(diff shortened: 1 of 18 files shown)


gnmi_server/server_test.go
```diff
@@ -3828,14 +3828,14 @@ print('%s')
...
-	initFullCountersDb(t, sdcfg.GetDbDefaultNamespace())
+	initFullConfigDb(t, sdcfg.GetDbDefaultInstance())
+	initFullCountersDb(t, sdcfg.GetDbDefaultInstance())
 
 	path, _ := os.Getwd()
 	path = filepath.Dir(path)
 
 	var cmd *exec.Cmd
-	cmd = exec.Command("bash", "-c", "cd "+path+" && "+"pytest")
+	cmd = exec.Command("bash", "-c", "cd "+path+" && "+"pytest -m 'not dpu'")
```
qiluo-msft, line 3838: > -m 'not dpu' [](http://example.com/codeflow?start=59&length=13)

Could you add some comment to explain the intention?

gnmi_server/server_test.go
```diff
@@ -3861,6 +3861,41 @@ print('%s')
 	s.s.Stop()
 }
 
+// Test DPU configuration with multiple databases
```
qiluo-msft, line 3864: >DPU [](http://example.com/codeflow?start=8&length=3)

This PR is gnmi general support the multi-db feature, not specific related to DPU, but also multi-asic, even single-asic with multiple database. Could you generalize naming/comments/etc ?

gnmi_server/server_test.go
ganglyu: updated

gnmi_server/server_test.go
ganglyu: updated

qiluo-msft: also generalize the PR ttile?

sonic_db_config/db_config.go
```diff
@@ -168,7 +185,8 @@ func GetDbTcpAddr(db_name string, ns string) string {
 	return hostname + ":" + strconv.Itoa(port)
 }
 
-func DbGetNamespaceAndConfigFile(ns_to_cfgfile_map map[string]string) {
+func DbParseConfigFile(name_to_cfgfile_map map[string]string) {
```
qiluo-msft, line 188: >DbParseConfigFile [](http://example.com/codeflow?start=5&length=17)

Let's explore using swss-common SonicDBConfig class instead of reimplementing in this file. We can raise another PR, and let this PR depending on that PR.

PR #177, opened 2023-12-22
Change MD5 to SHA256 algorithm

zbud-msft:
#### Why I did it

Semgrep complains about use of MD5 and how it not secure, so will use sha256 sum 

#### How I did it

Use SHA256 hash instead of MD5

#### How to verify it

Manual testing

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/telemetry/telemetry.go
+++ b/telemetry/telemetry.go
@@ -10,7 +10,7 @@ import (
 	"io/ioutil"
 	"strconv"
 	"time"
-
+	"os"
 	log "github.com/golang/glog"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
--- a/telemetry/telemetry.go
+++ b/telemetry/telemetry.go
@@ -3,8 +3,10 @@ package main
 import (
 	"crypto/tls"
 	"crypto/x509"
-	"crypto/md5"
+	"crypto/sha256"
+	"encoding/hex"
 	"flag"
+	"io"
 	"io/ioutil"
 	"strconv"
 	"time"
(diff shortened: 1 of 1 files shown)


telemetry/telemetry.go
```diff
@@ -112,11 +114,10 @@ func main() {
 			}
 			certificate, err = tls.LoadX509KeyPair(*serverCert, *serverKey)
 			if err != nil {
-				currentTime := time.Now().UTC()
-				log.Infof("Server Cert md5 checksum: %x at time %s", md5.Sum([]byte(*serverCert)), currentTime.String())
-				log.Infof("Server Key md5 checksum: %x at time %s", md5.Sum([]byte(*serverKey)), currentTime.String())
 				log.Exitf("could not load server key pair: %s", err)
 			}
+			validateSHA256(*serverCert)
+			validateSHA256(*serverKey)
```
ganglyu, line 120: Original implementation is: when tls.LoadX509KeyPair has error, we generate log with md5 checksum of file name.
New implementation is: when tls.LoadX509KeyPair has no error, we generate log with sha256 checksum of file content.
Can you explain why?

telemetry/telemetry.go
```diff
@@ -209,3 +210,19 @@ func getflag(name string) string {
 	})
 	return val
 }
+
+func validateSHA256(file string) {
```
FengPan-Frank, line 214: Seems this function is returning void, does main function need to break if validation is failed?

zbud-msft:
Opened new PR as now will use SHA512 checksum instead of 256 https://github.com/sonic-net/sonic-gnmi/pull/186

telemetry/telemetry.go
zbud-msft: Should be generating log with sha512 checksum when loadx509 fails. Fixed in https://github.com/sonic-net/sonic-gnmi/pull/186

telemetry/telemetry.go
zbud-msft: Called in block that will exit in new PR: https://github.com/sonic-net/sonic-gnmi/pull/186

PR #179, opened 2024-01-10
Add signal handler to close gnmi server

zbud-msft:
#### Why I did it

When process exits or killed unexpectedly, there is a possibility that gnmi server on declared port may still be open. When process received SIGTERM or SIGKILL, we can now stop gnmi server on port.

/var/log$ sudo netstat -nap | grep 50051
tcp6       0      0 :::50051                :::*                    LISTEN      24588/telemetry

Feb  5 08:34:42.363614 INFO telemetry#supervisord: telemetry E0205 08:34:42.358957      19 telemetry.go:176] Failed to create gNMI server: failed to open listener port 50051: listen tcp :50051: bind: address already in use

#### How I did it

Add signal handler via goroutine with channel and when syscalls are made, write to channel to close server.

#### How to verify it

UT

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/gnmi_server/server.go
+++ b/gnmi_server/server.go
@@ -183,7 +183,7 @@ func (srv *Server) Stop() {
 		log.Errorf("Stop() failed: not initialized")
 		return
 	}
-	srv.s.Stop()
+	s.Stop()
 }
 
 // Address returns the port the Server is listening to.
--- a/gnmi_server/server.go
+++ b/gnmi_server/server.go
@@ -177,6 +177,15 @@ func (srv *Server) Serve() error {
 	return srv.s.Serve(srv.lis)
 }
 
+func (srv *Server) Stop() {
+	s := srv.s
+	if s == nil {
+		log.Errorf("Stop() failed: not initialized")
+		return
+	}
+	srv.s.Stop()
+}
+
 // Address returns the port the Server is listening to.
 func (srv *Server) Address() string {
 	addr := srv.lis.Addr().String()
--- a/telemetry/telemetry.go
+++ b/telemetry/telemetry.go
@@ -8,7 +8,10 @@ import (
 	"io/ioutil"
 	"strconv"
 	"time"
-
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
 	log "github.com/golang/glog"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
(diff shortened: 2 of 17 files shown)


gnmi_server/server.go
```diff
@@ -177,6 +177,15 @@ func (srv *Server) Serve() error {
 	return srv.s.Serve(srv.lis)
 }
 
+func (srv *Server) Stop() {
+	s := srv.s
+	if s == nil {
+		log.Errorf("Stop() failed: not initialized")
+		return
+	}
+	srv.s.Stop()
```
qiluo-msft, line 186: >srv.s [](http://example.com/codeflow?start=1&length=5)

Is it just `s`?

qiluo-msft:
How to verify whether "port may still be open"? Do you want to add a sonic-mgmt testcase?

telemetry/telemetry.go
```diff
@@ -183,11 +186,34 @@ func main() {
 	if *withMasterArbitration {
 		s.ReqFromMaster = gnmi.ReqFromMasterEnabledMA
 	}
-	
-	log.V(1).Infof("Auth Modes: ", userAuth)
+
+	var reload = make(chan int, 1)
```
ganglyu, line 190: Why do you add this channel? Add some comments?

telemetry/telemetry.go
```diff
@@ -183,11 +186,34 @@ func main() {
...
+
+	wg.Add(1)
+
+	go signalHandler(reload, &wg, nil)
+
+	value := <-reload
+	log.V(1).Infof("Received Notification for gnmi server to shutdown")
+	s.Stop()
+	if value == 0 {
+		os.Exit(0)
```
ganglyu, line 213: If we exit here, will we lose syslog?

telemetry/telemetry.go
```diff
@@ -209,3 +235,17 @@ func getflag(name string) string {
...
 	return val
 }
+
+func signalHandler(reload chan<- int, wg *sync.WaitGroup, testSigChan <-chan os.Signal) {
+	defer wg.Done()
+	sigchannel := make(chan os.Signal, 1)
+	signal.Notify(sigchannel,
+		syscall.SIGTERM,
+		syscall.SIGQUIT)
+	select {
```
ganglyu, line 245: Can you explain sigchannel and testSigChan?

telemetry/telemetry_test.go
```diff
@@ -0,0 +1,37 @@
...
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+
+	go signalHandler(reload, wg, testSigChan)
+
+	testSigChan <- syscall.SIGTERM
+
+	select {
+	case val := <-reload:
+		if val != 0 {
```
ganglyu, line 28: Is it possible to get non-zero value here?

telemetry/telemetry_test.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"os"
+	"testing"
+	"context"
+	"time"
+	"sync"
+	"syscall"
+)
+
+func TestSignalHandler(t *testing.T) {
```
ganglyu, line 12: This test can cover SignalHandler, how do we test main?

zbud-msft: Main function now does not contain any real functional logic besides calling setup flags and goroutines

telemetry/telemetry_test.go
zbud-msft: As of right now, no, but with cert rotation in the future we will use a non-zero value to indicate that we should not exit goroutine but stop server, rotate certs, and restart server.

telemetry/telemetry.go
zbud-msft: Removed

telemetry/telemetry.go
zbud-msft: Removed

telemetry/telemetry.go
zbud-msft: Added comments, reload channel is used to as a way for gnmi server to stop. serve() is now executed in a goroutine and we use reload channel such that if it gets a value, we will stop server and exit.

gnmi_server/server.go
zbud-msft: Changed to just s.

zbud-msft:
> How to verify whether "port may still be open"? Do you want to add a sonic-mgmt testcase?

Will create a sonic-mgmt testcase

zbud-msft:
Depends on https://github.com/sonic-net/sonic-gnmi/pull/186 for fixing semgrep

PR #182, opened 2024-01-16
Go Code format checker and formatter

faraazbrcm:
Also formatted the files which dont adhere to go formats. Build will fail if there exists any formatting issue.
Formatting is supported for bookworm as well.

#### Why I did it

This was done to make the code readable and organised as per Golang standards

#### How I did it

I execute gofmt tool as part of build step and fail the build in case of formatting issues

#### How to verify it

build has to go through. no manual verification is required.


adyeung:
@kwangsuk @ganglyu please help review

Makefile
```diff
@@ -136,6 +141,22 @@ clean:
...
 
+# File target that generates a diff file if formatting is incorrect
+$(FORMAT_CHECK): $(GO_FILES)
+	@echo "Checking Go file formatting..."
+	@echo $(GO_FILES)
+	mkdir -p $(@D)
+	@$(GOROOT)/bin/gofmt -l -d $(GO_FILES) > $(FORMAT_LOG)
+	@if [ -s $(FORMAT_LOG) ]; then \
+		cat $(FORMAT_LOG); \
+		echo "Formatting issues found. Please run 'gofmt -w .' on the above files and commit the changes."; \
```
sachinholla, line 152: IMO this message will be confusing since the FORMAT_LOG file includes the diffs too. Better to remove `-d` option from the `gofmt` call so that FORMAT_LOG contains only the file names.

gnmi_server/clear_neighbor_dummy_test.go
```diff
@@ -0,0 +1,49 @@
+package gnmi
+
```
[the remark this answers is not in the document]
faraazbrcm: Corrected. Removed duplicate testcase.

Makefile
faraazbrcm: Done as suggested

sachinholla (approved):

kwangsuk (approved):

slicking:
@faraazbrcm thank you for putting up this PR, do you have the time to move it forward and submit?  We think the formatting cleanup and build check would be very helpful to have.

faraazbrcm:
> @faraazbrcm thank you for putting up this PR, do you have the time to move it forward and submit? We think the formatting cleanup and build check would be very helpful to have.

Its merged now

PR #183, opened 2024-01-19
[202205] Use ASICx suffix in DB target path only if the device is multi-asic

abdosi:
Cherry-pick of PR: https://github.com/sonic-net/sonic-gnmi/pull/184


abdosi:
Not needed for 202205 as having local patch.

PR #184, opened 2024-01-20
Use ASICx suffix in DB target path only if the device is multi-asic

abdosi:
What I did:
For supporting multi-asic as described here: https://github.com/sonic-net/sonic-telemetry/pull/77 we added support to handle `ASICx1` suffix in the target DB path. We  should only use that suffix  if the device is multi-asic else we can ignore it. This PR handle this case.

Why I did:
Telemetry Client can pass `ASICx` even for single asic device. To protect from that condition this change has been added.

How I verify:
Manually Verified . Before the fix client was getting error. After change it is getting data as expected.

### DIFF
--- a/sonic_data_client/db_client.go
+++ b/sonic_data_client/db_client.go
@@ -465,15 +465,14 @@ func IsTargetDb(target string) (string, bool, string, bool) {
 	dbName := targetname[0]
 	dbNameSpaceExist := false
 	dbNamespace, _ := sdcfg.GetDbDefaultNamespace()
-	isMultiNamespace := false
+	isMultiNamespace, _ := sdcfg.CheckDbMultiNamespace()
 
 	if len(targetname) > 2 {
 		log.V(1).Infof("target format is not correct")
 		return dbName, false, dbNamespace, dbNameSpaceExist
 	}
 	// ASIC Suffix is only used in case if device is multi-asic/namespace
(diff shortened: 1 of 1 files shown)


judyjoseph (approved):
LGTM, any unit tests needs a fix to cover this change ?

@judyjoseph @zbud-msft  will add sonic-mgmt test case for this. It will be hard to cover this via UT.

zbud-msft (approved):

sonic_data_client/db_client.go
```diff
@@ -465,13 +465,14 @@ func IsTargetDb(target string) (string, bool, string, bool) {
...
+	isMultiNamespace, _ := sdcfg.CheckDbMultiNamespace()
 
 	if len(targetname) > 2 {
 		log.V(1).Infof("target format is not correct")
 		return dbName, false, dbNamespace, dbNameSpaceExist
 	}
-
-	if len(targetname) > 1 {
+	// ASIC Suffix is only used in case if device is multi-asic/namespace
+	if isMultiNamespace && len(targetname) > 1 {
```
zbud-msft, line 475: "Before the fix client was getting error. After change it is getting data as expected." Can we get UT for this case?

yxieca:
@zbud-msft please add ADO number to this PR. thanks!

PR #186, opened 2024-02-01
Enable unit tests and code coverage for telemetry.go

zbud-msft:
#### Why I did it

telemetry.go currently is bypassing UT and diff code coverage checks.

What this PR is doing:

1. No longer uses a global config in order to make flag parsing and other goroutine functions easier to test by passing local config. Instead uses a local config struct TelemetryConfig for parsing and is created within setupFlags which parses flags passed by cmd line arguments. 
2. Logic for creating gnmi server and calling Serve() is now done inside of a goroutine such that it allows the main thread to support concurrent tasks in the future.
3. Call server stop after Serve() returns
4. MD5 checksum is replaced with SHA512 checksum as per semgrep findings. MD5 checksum is no longer considered secure as "MD5 hash algorithm is considered insecure. MD5 is not collision resistant and is therefore not suitable as a cryptographic signature."
5. Enable UT and code coverage for telemetry.go

#### How I did it

Removed global config, changed md5 checksum semgrep issue, enabled UT and CC checks for telemetry.go

#### How to verify it

Pipeline/UT

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


telemetry/telemetry.go
```diff
@@ -209,3 +210,19 @@ func getflag(name string) string {
 	})
 	return val
 }
+
+func validateSHA512(file string) {
```
ganglyu, line 214: This function performs two tasks: it computes the checksum of a given file using the SHA-512 algorithm and it logs a message to the system logger with the file name and the checksum value. Therefore, the function name validateSHA512 is misleading, as it implies that the function only validate checksum.

ganglyu:
How long does it take to calculate SHA512 checksum? Much slower?
It seems that our pr checker does not cover this file, would you please add unit test and check coverage?

telemetry/telemetry.go
zbud-msft: Changed

zbud-msft:
> How long does it take to calculate SHA512 checksum? Much slower? It seems that our pr checker does not cover this file, would you please add unit test and check coverage?

Added UT and code coverage for telemetry.go. Since we are computing SHA512 checksum of a cert file (~4KB for private key), we can expect it to take only a couple of MS.

ganglyu (dismissed):

ganglyu (approved):

PR #197, opened 2024-03-06
Update GNMI path schema

ganglyu:
#### Why I did it
We need to update sonic-gnmi to support new GNMI path schema.
HLD PR: https://github.com/sonic-net/SONiC/pull/1625

#### How I did it
I add a new element in GNMI path to support database instance.
```
/APPL_DB/localhost/DASH_VNET_TABLE/vnet1/
```
The second element is used for target database instance, and it's localhost by default.

#### How to verify it
Run unit test for sonic-gnmi.
Run end to end test for dash and gnmi.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/sonic_data_client/mixed_db_client.go
+++ b/sonic_data_client/mixed_db_client.go
@@ -275,7 +275,7 @@ func (c *MixedDbClient) gnmiFullPath(prefix, path *gnmipb.Path) (*gnmipb.Path, e
 			elems = append(prefix.GetElem(), elems...)
 		}
 		// Skip first two elem
-		// New GNMI path schema, /CONFIG_DB/localhost/PORT
+		// GNMI path schema is /CONFIG_DB/localhost/PORT
 		if len(elems) < 2 {
 			return nil, fmt.Errorf("Invalid gnmi path: length %d", len(elems))
 		}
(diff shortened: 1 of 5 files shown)


test/test_gnmi_appldb.py
```diff
@@ -8,16 +8,16 @@
 test_data_update_normal = [
     [
         {
-            'update_path': '/sonic-db:APPL_DB/DASH_QOS',
-            'get_path': '/sonic-db:APPL_DB/_DASH_QOS',
+            'update_path': '/sonic-db:APPL_DB/localhost/DASH_QOS',
```
wen587, line 11: Do we also use localhost in APPL_DB? I remember the localhost is only used in CONFIG_DB.

sonic_data_client/mixed_db_client.go
```diff
@@ -269,23 +269,19 @@ func (c *MixedDbClient) gnmiFullPath(prefix, path *gnmipb.Path) *gnmipb.Path {
...
-	}
 	if path.GetElem() != nil {
 		elems := path.GetElem()
 		if prefix != nil {
 			elems = append(prefix.GetElem(), elems...)
 		}
-		// Skip first elem
-		fullPath.Elem = elems[1:]
+		// Skip first two elem
+		// New GNMI path schema, /CONFIG_DB/localhost/PORT
```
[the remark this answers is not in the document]
ganglyu: Fixed.

test/test_gnmi_appldb.py
ganglyu: GNMI needs to support multi-asic for APPL_DB, so I use the same schema for APPL_DB.

wen587 (approved):

qiluo-msft (approved):

PR #198, opened 2024-03-07
Call flag.Parse() to parse global flags like -logtostderr

zbud-msft:
#### Why I did it

Currently we are using a custom flagset, however glog requires that it's flags like logtostderr and v, be part of the global flag set. I parse cmd line args such that we call flag.Parse on only the glog global flags and then call the custom flagset parse on non glog flags.

#### How I did it

Parse os.args such that we separate it into glog flags and telemetry flags. Call flag.Parse() to set up the global flags that glog requires, then flagset.Parse() to set up the flags that telemetry config will use.

#### How to verify it

UT

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/telemetry/telemetry.go
+++ b/telemetry/telemetry.go
@@ -85,24 +85,25 @@ func runTelemetry(args []string) error {
 	return nil
 }
 
-func getGlogPrefixes() []string {
+func getGlogsFlagMap() []string {
 	// glog flags: https://pkg.go.dev/github.com/golang/glog
-	return []string {
-		"-alsologtostderr",
-		"-log_backtrace_at",
-		"-log_dir",
-		"-log_link",
-		"-logbuflevel",
-		"-logtostderr",
-		"-stderrthreshold",
-		"-v",
-		"-vmodule",
+	return map[string]bool {
+		"-alsologtostderr":  true,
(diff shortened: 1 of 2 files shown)


telemetry/telemetry.go
```diff
@@ -72,6 +85,46 @@ func runTelemetry(args []string) error {
...
+	telemetryFlags := []string{os.Args[0]}
+
+	for _, arg := range os.Args[1:] {
+		if strings.HasPrefix(arg, "-v") { // only flag in both glog and telemetry
+			glogFlags = append(glogFlags, arg)
+			telemetryFlags = append(telemetryFlags, arg)
+			continue
+		}
+		isGlogFlag := false
+		for _, prefix := range getGlogPrefixes() {
```
ganglyu, line 114: Can you use hashmap to avoid iterating list?

zbud-msft: Done

qiluo-msft:
Could you check unit test failure?

zbud-msft:
> Could you check unit test failure?

Done

ganglyu (approved):

PR #199, opened 2024-03-07
[202305] Use ASICx suffix in DB target path only if the device is multi-asic

abdosi:
Cherry-pick of PR: https://github.com/sonic-net/sonic-gnmi/pull/184


zbud-msft (approved):

StormLiangMS (approved):
LGTM

StormLiangMS:
@abdosi could you run test with 202305 image to avoid regression at this late stage ?

PR #203, opened 2024-03-13
Add inotify monitoring for certificates

zbud-msft:
#### Why I did it

Telemetry Process will now no longer exit when certs are missing. Because we added a 
goroutine within startGNMIServer such that we create a fsnotify watcher on the certificates directory to monitor changes brought upon files within that directory, such as deletion, creation, and modification, we will wait for certs to be present and then start gnmi server to start serving.

GNMI server will now gracefully stop such that new requests are no longer accepted but pending requests will be finished, and then server will stop and process will exit.

Once gnmi server has started serving, if we see fsnotify events (create, write, remove, rename), that means the certificate files within the monitored directory has been modified/created, and we will restart gnmi server with new certificates or if someone has maliciously deleted cert, we will until certs are present.

If there are errors, that will not be fixed by retrying or new certs, we will stop gnmi server. (Goal would be to graceful stop, I will do that change in another PR)

#### How I did it

- Added fsnotify to go.mod
- runTelemetry will wait on startGNMIServer goroutine and start another goroutine signalHandler to monitor for syscalls such as SIGTERM or SIGQUIT
- signalHandler will read from signchannel that is notified when SIGTERM or SIGQUIT, and will then write to a channel that startGNMIServer will read from to stop.
- startGNMIServer is now looped such that it will spawn a cert monitoring goroutine in the case where the telemetry config is TLS and not insecure every iteration.
- startGNMIServer will try to load certs and if it unable to do so, corrupted or missing certs, will wait for an fsnotify write or create that cert monitoring will send. We will try to load certs again after .create or .write. Once successfully loaded certs, we will start gnmi server. 
- Once gnmi server has started serving, if we see a fsnotify event (remove, create, write, or rename) we will stop the server (graceful stop in the future) and then finish the current iteration of the loop in which we will try loading new certs again or wait for certs to be present.

iNotifyCertMonitoring goroutine will close after every time it receives a fsnotify event/sends a signal to startGNMIServer so that we do not worry about leak.

#### How to verify it

UT

I will also add sonic-mgmt testcases.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/gnmi_server/server.go
+++ b/gnmi_server/server.go
@@ -56,7 +56,7 @@ type Config struct {
 	UserAuth AuthTypes
 	EnableTranslibWrite bool
 	EnableNativeWrite bool
-	ZmqAddress string
+	ZmqPort string
 	IdleConnDuration int
 }
 
@@ -367,7 +367,7 @@ func (s *Server) Get(ctx context.Context, req *gnmipb.GetRequest) (*gnmipb.GetRe
 			}
 		}
 		if check := IsNativeOrigin(origin); check {
-			dc, err = sdc.NewMixedDbClient(paths, prefix, origin, encoding, s.config.ZmqAddress)
+			dc, err = sdc.NewMixedDbClient(paths, prefix, origin, encoding, s.config.ZmqPort)
 		} else {
 			dc, err = sdc.NewTranslClient(prefix, paths, ctx, extensions)
 		}
@@ -446,7 +446,7 @@ func (s *Server) Set(ctx context.Context, req *gnmipb.SetRequest) (*gnmipb.SetRe
 			common_utils.IncCounter(common_utils.GNMI_SET_FAIL)
 			return nil, grpc.Errorf(codes.Unimplemented, "GNMI native write is disabled")
 		}
-		dc, err = sdc.NewMixedDbClient(paths, prefix, origin, encoding, s.config.ZmqAddress)
+		dc, err = sdc.NewMixedDbClient(paths, prefix, origin, encoding, s.config.ZmqPort)
 	} else {
 		if s.config.EnableTranslibWrite == false {
 			common_utils.IncCounter(common_utils.GNMI_SET_FAIL)
@@ -521,7 +521,7 @@ func (s *Server) Capabilities(ctx context.Context, req *gnmipb.CapabilityRequest
 	var supportedModels []gnmipb.ModelData
 	dc, _ := sdc.NewTranslClient(nil, nil, ctx, extensions)
 	supportedModels = append(supportedModels, dc.Capabilities()...)
-	dc, _ = sdc.NewMixedDbClient(nil, nil, "", gnmipb.Encoding_JSON_IETF, s.config.ZmqAddress)
+	dc, _ = sdc.NewMixedDbClient(nil, nil, "", gnmipb.Encoding_JSON_IETF, s.config.ZmqPort)
 	supportedModels = append(supportedModels, dc.Capabilities()...)
 
 	suppModels := make([]*gnmipb.ModelData, len(supportedModels))
--- a/gnmi_server/server.go
+++ b/gnmi_server/server.go
@@ -177,13 +177,13 @@ func (srv *Server) Serve() error {
 	return srv.s.Serve(srv.lis)
 }
 
-func (srv *Server) Stop() {
+func (srv *Server) GracefulStop() {
 	s := srv.s
 	if s == nil {
 		log.Errorf("Stop() failed: not initialized")
 		return
 	}
-	s.Stop()
+	s.GracefulStop()
 }
 
 // Address returns the port the Server is listening to.
--- a/gnmi_server/server.go
+++ b/gnmi_server/server.go
@@ -177,13 +177,13 @@ func (srv *Server) Serve() error {
 	return srv.s.Serve(srv.lis)
 }
 
-func (srv *Server) GracefulStop() {
+func (srv *Server) Stop() {
 	s := srv.s
 	if s == nil {
 		log.Errorf("Stop() failed: not initialized")
 		return
 	}
-	s.GracefulStop()
+	s.Stop()
 }
 
 // Address returns the port the Server is listening to.
--- a/telemetry/telemetry.go
+++ b/telemetry/telemetry.go
@@ -42,6 +42,7 @@ type TelemetryConfig struct {
 	ServerCert            *string
 	ServerKey             *string
 	ZmqAddress            *string
+	ZmqPort               *string
 	Insecure              *bool
 	NoTLS                 *bool
 	AllowNoClientCert     *bool
@@ -146,7 +147,8 @@ func setupFlags(fs *flag.FlagSet) (*TelemetryConfig, *gnmi.Config, error) {
 		CaCert:                fs.String("ca_crt", "", "CA certificate for client certificate validation. Optional."),
 		ServerCert:            fs.String("server_crt", "", "TLS server certificate"),
 		ServerKey:             fs.String("server_key", "", "TLS server private key"),
-		ZmqAddress:            fs.String("zmq_address", "", "Orchagent ZMQ address, when not set or empty string telemetry server will switch to Redis based communication channel."),
+		ZmqAddress:            fs.String("zmq_address", "", "Orchagent ZMQ address, deprecated, please use zmq_port."),
+		ZmqPort:               fs.String("zmq_port", "", "Orchagent ZMQ port, when not set or empty string telemetry server will switch to Redis based communication channel."),
 		Insecure:              fs.Bool("insecure", false, "Skip providing TLS cert and key, for testing only!"),
 		NoTLS:                 fs.Bool("noTLS", false, "disable TLS, for testing only!"),
 		AllowNoClientCert:     fs.Bool("allow_no_client_auth", false, "When set, telemetry server will request but not require a client certificate."),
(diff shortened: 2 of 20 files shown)


telemetry/telemetry.go
```diff
@@ -219,113 +272,127 @@ func signalHandler(serverControlSignal chan<- int, wg *sync.WaitGroup, sigchanne
 	}
 }
 
-func startGNMIServer(telemetryCfg *TelemetryConfig, cfg *gnmi.Config, serverControlSignal <-chan int, wg *sync.WaitGroup) {
```
qiluo-msft, line 222: ><-chan int [](http://example.com/codeflow?start=90&length=10)

Do you really need to change type? I see iNotifyCertMonitoring requires old type.

telemetry/telemetry_test.go
```diff
@@ -197,6 +205,14 @@ func saveCertKeyPair(certPath, keyPath string) error {
 	certBytes := cert.Certificate[0]
 	keyBytes := x509.MarshalPKCS1PrivateKey(cert.PrivateKey.(*rsa.PrivateKey))
 
+	// Create the parent certs directory
+
+	dirPath := filepath.Dir(certPath)
+
+	if err := os.MkdirAll(dirPath, 0755); err != nil {
```
qiluo-msft, line 212: >MkdirAll [](http://example.com/codeflow?start=14&length=8)

Could you prevent create dir inside test? How about just create the dir in codebase?

zbud-msft: Sure, will change.

telemetry/telemetry.go
zbud-msft: Yes, we will need to use a bidirectional channel here, since in StartGNMIServer goroutine we will be reading from serverControlSignal to stop/restart and we will also spawn monitoring goroutine which will be writing to serverControlSignal when cert files are deleted/modified.

telemetry/telemetry.go
```diff
@@ -211,6 +213,57 @@ func isFlagPassed(fs *flag.FlagSet, name string) bool {
...
+			testReadySignal <- 0
+		}
+		for {
+			select {
+			case event := <-watcher.Events:
+				if event.Name != "" {
+					log.V(1).Infof("Inotify watcher has received event: %v", event)
+					if event.Op & fsnotify.Write == fsnotify.Write {
+						log.V(1).Infof("Cert File has been modified: %s", event.Name)
+						serverControlSignal <- 1 // tells gnmi server to stop and restart with new certs
```
qiluo-msft, line 231: >1 [](http://example.com/codeflow?start=29&length=1)

magic numbers. In future if more enum added, it's not easy to understand the code.

telemetry/telemetry.go
```diff
@@ -211,6 +213,57 @@ func isFlagPassed(fs *flag.FlagSet, name string) bool {
...
+						done <- true
+						return
+					}
+					if event.Op & fsnotify.Remove == fsnotify.Remove {
+						log.Errorf("Cert file has been deleted: %s", event.Name)
+						serverControlSignal <- 0 // tells gnmi server to stop
```
qiluo-msft, line 237: >stop [](http://example.com/codeflow?start=55&length=4)

Is it force stopping or graceful stopping in your design?

qiluo-msft: We may design both, but implement at least the force stopping if that is minimal required.

telemetry/telemetry.go
```diff
@@ -211,6 +213,57 @@ func isFlagPassed(fs *flag.FlagSet, name string) bool {
...
+			case event := <-watcher.Events:
+				if event.Name != "" {
+					log.V(1).Infof("Inotify watcher has received event: %v", event)
+					if event.Op & fsnotify.Write == fsnotify.Write {
+						log.V(1).Infof("Cert File has been modified: %s", event.Name)
+						serverControlSignal <- 1 // tells gnmi server to stop and restart with new certs
+						done <- true
+						return
+					}
+					if event.Op & fsnotify.Remove == fsnotify.Remove {
```
qiluo-msft, line 235: >Remove [](http://example.com/codeflow?start=47&length=6)

Is it possible that a cert file overwrite generate Create event immediate after Remove event? If it is the behavior, you have no way to start.

telemetry/telemetry.go
zbud-msft: ServerControlSignal channel receives a value when SIGKILL or SIGTERM is received, monitored cert file has been deleted or changed, cert watcher has received a non-filesystem error.

When serverControlSignal channel receives a value, the running gnmi server will always be force stopped via s.Stop(). And depending on the value, gnmi server will either be restarted or not. If cert is rotated, we will restart gnmi server, but in the other cases we will let the goroutine return.

https://pkg.go.dev/google.golang.org/grpc#Server.Stop

"Stop stops the gRPC server. It immediately closes all open connections and listeners. It cancels all active RPCs on the server side and the corresponding pending RPCs on the client side will get notified by connection errors."

telemetry/telemetry.go
zbud-msft: Will use enums to make code easier to understand.

testdata/certs/README.md
```diff
@@ -0,0 +1 @@
+This empty directory will be used by telemetry_test.go to add certs to.
```
qiluo-msft, line 1: >empty [](http://example.com/codeflow?start=5&length=5)

Your solution works. However, there is a community fancier implementation: [https://stackoverflow.com/a/932982](https://stackoverflow.com/a/932982)

telemetry/telemetry.go
qiluo-msft: To suicide or not is a critical decision, and I do not see how can we prove 1 second is a good divider for the decision. Let's assume the time is 2 seconds, the gnmi process is already suicided, will it auto restart?

telemetry/telemetry.go
qiluo-msft: so it is force stop.

telemetry/telemetry.go
qiluo-msft: Is it a good strategy that if cert is missing, stop gnmi serving any new incoming request, and silently wait for a new cert?

zbud-msft: Implement logic such that

- If certs are missing when telemetry process is started, we will wait for fsnotify.Create or fsnotify.Write event to check for certs again. Only when certs are present and are good, do we start gnmi server to start serving.

- If the case where certs are overwritten, or are missing as gnmi server is running, we will gracefully stop gnmi server, and then try loading the new certs or wait for certs.

testdata/certs/README.md
zbud-msft: Done.

qiluo-msft:
@ganglyu Could you also help review?

qiluo-msft (approved):
LGTM. Wait for more reviewers.

gnmi_server/server.go
```diff
@@ -177,13 +177,13 @@ func (srv *Server) Serve() error {
 	return srv.s.Serve(srv.lis)
 }
 
-func (srv *Server) Stop() {
+func (srv *Server) GracefulStop() {
```
ganglyu, line 180: Do we really need to rename this API?

gnmi_server/server.go
```diff
@@ -177,13 +177,13 @@ func (srv *Server) Serve() error {
...
 	s := srv.s
 	if s == nil {
 		log.Errorf("Stop() failed: not initialized")
```
ganglyu, line 183: Log is still using Stop()

telemetry/telemetry.go
```diff
@@ -70,16 +80,16 @@ func runTelemetry(args []string) error {
...
-	// serverControlSignal channel is a channel that will be used to notify gnmi server of sigcall which should stop server
-	var serverControlSignal = make(chan int, 1)
+	// serverControlSignal channel is a channel that will be used to notify gnmi server to start, stop, restart, depending of syscall or cert updates
+	var serverControlSignal = make(chan ServerControlValue, 1)
 	sigchannel := make(chan os.Signal, 1)
 	signal.Notify(sigchannel, syscall.SIGTERM, syscall.SIGQUIT)
 
-	wg.Add(1)
-	go startGNMIServer(telemetryCfg, cfg, serverControlSignal, &wg)
+	go signalHandler(serverControlSignal, sigchannel)
```
ganglyu, line 88: Why do we remove wg from signalHandler?

telemetry/telemetry.go
```diff
@@ -211,121 +229,207 @@ func isFlagPassed(fs *flag.FlagSet, name string) bool {
...
+
+	go func() {
+		if testReadySignal != nil { // for testing only
+			testReadySignal <- 0
+		}
+		for {
+			select {
+			case event := <-watcher.Events:
+				if event.Name != "" && (filepath.Ext(event.Name) == ".cert" || filepath.Ext(event.Name) == ".crt" ||
+                                   filepath.Ext(event.Name) == ".pem" || filepath.Ext(event.Name) == ".key") {
```
ganglyu, line 245: This line is using space instead of tab?

telemetry/telemetry.go
```diff
@@ -211,121 +229,207 @@ func isFlagPassed(fs *flag.FlagSet, name string) bool {
...
+				}
+			}
+		}
+	}()
+
+	telemetryCertDirectory := filepath.Dir(*telemetryCfg.ServerCert)
+
+	log.V(1).Infof("Begin cert monitoring on %s", telemetryCertDirectory)
+
+	err := watcher.Add(telemetryCertDirectory) // Adding watcher to cert directory
```
ganglyu, line 275: I think we don't need monitor the whole directory, and we only care about CA cert and chained cert.

telemetry/telemetry.go
zbud-msft: I will add change to gracefully stop gnmi server in a different PR.

gnmi_server/server.go
zbud-msft: Won't use Graceful stop in this PR, but yes, we don't have to change the name of the wrapper API

gnmi_server/server.go
zbud-msft: Will keep name

telemetry/telemetry.go
zbud-msft: Initially, I thought it doesn't matter if signalHandler is in the wg, since if startGNMIServer is the critical goroutine, if that exits we can have the process exit, but I think it is fine to add it to the wg to make sure all goroutines exit cleanly, avoid any potential issues or leaks.

zbud-msft: Added back to wg

telemetry/telemetry.go
zbud-msft: Fixed

telemetry/telemetry.go
zbud-msft: fsnotify recommends to not use watcher on specific files but on directories. I think it should be fine as we filter out the file events based off extension that we care about like .pem, .key, .cert, etc.

https://pkg.go.dev/github.com/fsnotify/fsnotify#section-readme

ganglyu (dismissed):
LGTM

qiluo-msft (approved):

qiluo-msft (approved):

PR #207, opened 2024-03-29
Let pfcwdmap creation fail gracefully

zbud-msft:
ADO: 27075747
#### Why I did it

There is a chance that PORT_QOS_MAP|global is used to fetc pfc_enable key, which will cause redis hget to fail and return err. All queries to counters_db will fail after that.

- Fix that |global will not break redis hget
- Failure to create pfcwd map will not impact COUNTERS_DB queries. 

#### How I did it

Iterate and not use global as key for port_qos_map to check for pfc_enable

Any error creating pfcwd map will not stop future COUNTERS_DB queries

#### How to verify it

UT/Manual test. Tested in 202305

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/sonic_data_client/virtual_db.go
+++ b/sonic_data_client/virtual_db.go
@@ -110,7 +110,7 @@ func initAliasMap() error {
 func initCountersPfcwdNameMap() error {
 	var err error
 	if len(countersPfcwdNameMap) == 0 {
-		countersPfcwdNameMap, err = getPfcwdMap()
+		countersPfcwdNameMap, err = GetPfcwdMap()
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


sonic_data_client/virtual_db.go
```diff
@@ -166,7 +166,14 @@ func getPfcwdMap() (map[string]map[string]string, error) {
 			log.V(1).Infof("PFC WD not enabled on device")
 			return nil, nil
 		}
-		qos_key := resp[0]
+
+		var qos_key string
+		for _, key := range resp {
+			if strings.Contains(key, "Ethernet") { // Account for PORT_QOS_MAP|global
```
qiluo-msft, line 172: >global [](http://example.com/codeflow?start=70&length=6)

could you just check "global" in the loop?

qiluo-msft: I understand this is to filter all key containing "Ethernet". However this prefix may be changed in future. Check not equal to "global" may be safer.

rlucus (approved):

PR #178, opened 2024-01-09
Account for GLOBAL key in PFC_WD

zbud-msft:
#### Why I did it

In some cases in PFC_WD table there are keys such as "PFC_WD|GLOBAL" and GLOBAL cannot be mapped to COUNTERS_QUEUE_NAME_MAP so we run into an error since we cannot retrieve counters for key with not interface declared.  

#### How I did it

Omit keys from PFC_WD that do not declare an interface

#### How to verify it

UT/manaul testing

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

PR #208, opened 2024-04-02
Revert "[action] [PR:178] Account for GLOBAL key in PFC_WD (#178)"

zbud-msft:
Reverts sonic-net/sonic-gnmi#200


zbud-msft:
In favor of https://github.com/sonic-net/sonic-gnmi/pull/209

PR #213, opened 2024-04-08
Add KillProcess gNOI API

isabelmsft:
#### Why I did it
Add gNOI API support for stop and restart process. 

With this addition, an external client with write permissions can request a process restart or stop. In this first phase, only a subset of services are supported: ['snmp', 'swss', 'dhcp_relay', 'radv', 'restapi', 'lldp', 'sshd']. sonic-host-services dbus rejects any other service management requests.

#### How I did it

- Upgrade gNOI version to include KillProcess API
- Update gNOI client to support KillProcess rpc
- Implement gNOI API to parse necessary arguments (service name and restart request boolean) from client and invoke dbus API based on client request
- Support KillProcess and service restart through dbus in sonic-host-services, added in this PR: https://github.com/sonic-net/sonic-host-services/pull/111/files 

#### How to verify it
Unit tests and E2E tests

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


qiluo-msft:
Could you add this PR link into HLD PR description? add a code PR table there.

gnmi_server/gnoi.go
```diff
@@ -20,6 +20,40 @@ import (
...
+		if err != nil {
+			log.V(2).Infof("Failed to stop service %s: %v", serviceName, err)
+		}
+	}
+	return err
+}
+
+func (srv *Server) KillProcess(ctx context.Context, req *gnoi_system_pb.KillProcessRequest) (*gnoi_system_pb.KillProcessResponse, error) {
+	serviceName := req.GetName()
+	restart := req.GetRestart()
```
qiluo-msft, line 46: >GetRestart [](http://example.com/codeflow?start=16&length=10)

I see other useful fields like `signal`. Could you check if we need to parse them all? We can implement only limited fields, and return NotImplemented for other combination.

gnmi_server/gnoi.go
```diff
@@ -20,6 +20,40 @@ import (
...
+		err = sc.StopService(serviceName)
+		if err != nil {
+			log.V(2).Infof("Failed to stop service %s: %v", serviceName, err)
+		}
+	}
+	return err
+}
+
+func (srv *Server) KillProcess(ctx context.Context, req *gnoi_system_pb.KillProcessRequest) (*gnoi_system_pb.KillProcessResponse, error) {
+	serviceName := req.GetName()
```
qiluo-msft, line 45: >serviceName [](http://example.com/codeflow?start=1&length=11)

Where is the code to check serviceName is allowed? For example, kill or restart gnmi or syncd is problematic.

test/utils.py
```diff
@@ -188,6 +188,29 @@ def gnoi_reboot(method, delay, message):
...
+    cmd = path + '/build/bin/gnoi_client '
+    cmd += '-insecure -target 127.0.0.1:8080 '
+    cmd += '-rpc KillProcess '
+    cmd += f'-jsonin \'{json_data}\''
+    ret, msg = run_cmd(cmd)
+    return ret, msg
+
+def gnoi_restart_process(valid):
+    if not valid:
+        json_data = '{"name": "snmp", "restart": invalid}'
```
FengPan-Frank, line 203: Does invalid exist in boolean? or you may want ot use string instead.

isabelmsft: invalid does not exist in boolean, but this is meant to test a failure scenario with invalid option passed. For example, this is the expected outcome when running gnoi_client with an invalid option passed:

```
root@str-s6100-acs-1:/# gnoi_client -target 127.0.0.1:50052 -logtostderr -insecure -rpc KillProcess -jsonin '{"name": "snmp", "restart": invalid}'    
Kill Process with optional restart
panic: invalid character 'i' looking for beginning of value

goroutine 1 [running]:
main.killProcess(0xb4cbc0, 0xc00027e150, 0xb47980, 0xc00007f240)
        /sonic/src/sonic-gnmi/gnoi_client/gnoi_client.go:120 +0x2b4
main.main()
        /sonic/src/sonic-gnmi/gnoi_client/gnoi_client.go:61 +0x7dd
```

gnmi_server/gnoi.go
isabelmsft: GNOI protocol defines fields that may be helpful, but there is no requirement to parse them all. I've added additional processing to output Unimplemented if one of the extraneous fields is passed (`signal` or `pid`) by the client.

gnmi_server/gnoi.go
isabelmsft: Checking if serviceName is allowed is done in dbus: https://github.com/sonic-net/sonic-host-services/blob/fa6fc0f2972a8ad10ca3a2dfcb8e9e93f5c24ad8/host_modules/systemd_service.py#L20

This error message is passed to gnmi_client. This scenario is tested in sonic-mgmt through this PR: https://github.com/sonic-net/sonic-mgmt/pull/12478/files

gnmi_server/gnoi.go
```diff
@@ -20,6 +20,46 @@ import (
...
+	restart := req.GetRestart()
+        if req.GetPid() != 0 {
+            return nil, status.Errorf(codes.Unimplemented, "Pid option is not implemented")
+        }
+        if req.GetSignal() != gnoi_system_pb.KillProcessRequest_SIGNAL_UNSPECIFIED {
+            return nil, status.Errorf(codes.Unimplemented, "Signal option is not implemented")
+        }
+	log.V(1).Info("gNOI: KillProcess with optional restart")
+	log.V(1).Info("Request:", req)
```
ganglyu, line 54: Does this log work? Do you have sample output?

gnmi_server/server.go
```diff
@@ -43,7 +43,9 @@ type Server struct {
 	// comes from a master controller.
 	ReqFromMaster func(req *gnmipb.SetRequest, masterEID *uint128) error
 	masterEID     uint128
+	gnoi_system_pb.UnimplementedSystemServer
```
ganglyu, line 46: what does this mean?

test/utils.py
```diff
@@ -188,6 +188,25 @@ def gnoi_reboot(method, delay, message):
     ret, msg = run_cmd(cmd)
     return ret, msg
 
+def gnoi_kill_process():
+    path = os.getcwd()
+    json_data = '{"name": "snmp"}'
```
ganglyu, line 193: Why do you use fixed data for kill process? I think gnoi_kill_process and gnoi_restart_process should be the same.

gnmi_server/gnoi.go
isabelmsft: Yes.

Sample log:
```
2024 May 22 01:31:51.650270 str-s6100-acs-1 INFO gnmi#supervisord: gnmi-native I0522 01:31:51.649487      27 gnoi.go:151] gNOI: Time
2024 May 22 01:32:17.224515 str-s6100-acs-1 INFO gnmi#supervisord: gnmi-native I0522 01:32:17.223930      27 gnoi.go:47] gNOI: KillProcess with optional restart
2024 May 22 01:32:17.224748 str-s6100-acs-1 INFO gnmi#supervisord: gnmi-native I0522 01:32:17.223964      27 gnoi.go:48] Request:name:"swss"  restart:true
2024 May 22 01:32:17.224920 str-s6100-acs-1 INFO gnmi#supervisord: gnmi-native I0522 01:32:17.224042      27 gnoi.go:29] Restarting service swss...
```

gnmi_server/server.go
isabelmsft: To satisfy SystemServer interface requirements for this gNOI version, all implementations must embed UnimplementedSystemServer (added to gNOI for forward compatibility. For reference: https://github.com/openconfig/gnoi/blob/349b3dcd04ecf78d740a25f17f983a8c8cd56e73/system/system_grpc.pb.go#L226)

ganglyu: Please add comment here

test/utils.py
isabelmsft: Thanks, changed to match gnoi_restart_process

gnmi_server/server.go
isabelmsft: Added comment

ganglyu (approved):
LGTM

gnmi_server/gnoi.go
```diff
@@ -20,6 +20,46 @@ import (
...
+		log.V(2).Infof("Stopping service %s...", serviceName)
+		err = sc.StopService(serviceName)
+		if err != nil {
+			log.V(2).Infof("Failed to stop service %s: %v", serviceName, err)
+		}
+	}
+	return err
+}
+
+func (srv *Server) KillProcess(ctx context.Context, req *gnoi_system_pb.KillProcessRequest) (*gnoi_system_pb.KillProcessResponse, error) {
```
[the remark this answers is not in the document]
isabelmsft: Thanks, I added auth, which uses ctx.

zbud-msft (dismissed):

gnmi_server/gnoi.go
qiluo-msft: let's treat no signal same as default signal (SIGTERM)

isabelmsft: Discussed offline, the agreed-upon implementation is to require the client to specify signal, as the protocol defines an unspecified signal as the "invalid default" ref: https://github.com/openconfig/gnoi/blob/349b3dcd04ecf78d740a25f17f983a8c8cd56e73/system/system.pb.go#L243

Because our KillProcess implementation uses generic `systemctl stop` and `restart` commands (which use SIGNAL_TERM signal by default), we require the client to specify the signal option 1: SIGNAL_TERM

qiluo-msft (approved):

PR #216, opened 2024-04-11
Bookworm migration for all dependency packages in azurepipeline for sonic-gnmi

ranjinidn:
#### Why I did it
As part of the migration of sonic-gnmi repo to bookworm, migrating the azure pipeline files to refer to the bookworm build for dependency installation.

#### How I did it
Changed all references from bullseye to bookworm in azure-pipeline for dependency packages installation.

#### How to verify it
The tests in azurepipeline were run on the bookworm build locally and has the same behavior as bullseye build.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


kwangsuk (approved):

kwangsuk:
@sachinholla please review.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sonic-net/sonic-gnmi
 
-go 1.15
+go 1.19
```
sachinholla, line 3: There is a concept of FIPS enabled build in sonic community and that build still uses go1.15. See https://github.com/sonic-net/sonic-buildimage/blob/master/sonic-slave-bookworm/Dockerfile.j2#L478. This change will break FIPS builds. I am not sure how to trigger one; but you can try setting INCLUDE_FIPS=y in `rules/config` file.

hi @qiluo-msft -- can you confirm whether INCLUDE_FIPS setting is being used by MSFT or someone? Bookworm slave docker installs go1.15 when INCLUDE_FIPS=y and go1.19 otherwise. This will break compilation of INCLUDE_FIPS =y builds if we use go1.19 specific libraries/syntax in the code -- and it won't be caught during PR sanity runs. The slave docker must always stick to one go version. I had brought it up in the [UMF WG meeting](https://lists.sonicfoundation.dev/g/umf-working-group/message/9) earlier but looks like there was no further updates. It would be great if you can clarify why two versions are being used and confirm if it is okay to ignore go1.15.

ranjinidn: The go version for bookworm build is already upgraded to 1.19. 
One of the most notable changes in Go 1.19 is the addition of support for FIPS 140-3.
Hence it makes sense to go ahead and use go 1.19 for FIPS as well.

We think we can keep the bookworm migration for sonic-gnmi apart from the FIPS migration to go1.19 separate.

@xumia, we see that the FIPS support was introduced by you in bullseye. (https://github.com/sonic-net/sonic-buildimage/commit/8ec8900d31806d4f641c91b2b62ec5880b504687)
Can you please take care of upgrading the FIPS to go1.19 as part of bookworm migration? 
We see some documentation about setting the environment variable GOEXPERIMENT=boringcrypto. This will cause the Go compiler to use BoringCrypto, a FIPS 140-3 validated cryptographic module.

sachinholla: > We think we can keep the bookworm migration for sonic-gnmi apart from the FIPS migration to go1.19 separate. Support for FIPS is disabled in bookworm build currently

Ok.. hopefully FIPS team too will upgrade to go1.19 and not revert this PR :)

sachinholla (approved):

ayelrod (approved):

go.mod
saiarcot895: The SONiC FIPS support for Bookworm appears to be in sonic-net/sonic-buildimage#18088, this looks like it'll provide a FIPS-enabled go1.19 package.

ranjinidn: @saiarcot895 Since the FIPS build is disabled for bookworm by default, can we merge this without waiting for the PR #18088 to be merged ?
Merging this gnmi PR will also make sure the sonic-gnmi repo is already migrated to go1.19 before the #18088 merges.
Also this is a dependency for https://github.com/sonic-net/sonic-buildimage/pull/18548.

saiarcot895 (approved):

PR #219, opened 2024-04-23
Use server GracefulStop instead of force stop

zbud-msft:
#### Why I did it

It is better for server to gracefully stop after all pending requests are finished while blocking incoming requests instead of force closing all requests.

#### How I did it

Use GracefulStop API instead of Stop

#### How to verify it

UT, will add sonic-mgmt testcase

Verified via manual test that if during server is serving request, if certs are modified, server will not force close request, but will in fact finish the request and then server will be closed and that no new request is allowed. Will create a sonic-mgmt testcase for this.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


gnmi_server/server.go
```diff
@@ -183,7 +183,7 @@ func (srv *Server) Stop() {
 		log.Errorf("Stop() failed: not initialized")
 		return
 	}
-	s.Stop()
+	s.GracefulStop()
```
qiluo-msft, line 186: >GracefulStop [](http://example.com/codeflow?start=3&length=12)

Is it true that `docker stop` will also trigger a graceful stop? If yes, I think we need to treat all the scenarios differently, only gracefully restart if rotating a cert.

zbud-msft: Yes that makes sense, docker stop will initially send a SIGTERM, if server has not stopped within a grace period (10s), docker stop will then send a SIGKILL which will force stop telemetry process. In most cases with this change, we will wait the 10s for the SIGKILL to come. I will change logic that if sigterm/sigint/sigquit/sighup is received we will call Stop() and not GracefulStop() which will be used only when rotating certs.

rlucus (approved):

telemetry/telemetry.go
```diff
@@ -464,12 +464,13 @@ func startGNMIServer(telemetryCfg *TelemetryConfig, cfg *gnmi.Config, serverCont
...
 		serverControlValue := <-serverControlSignal
 		log.V(1).Infof("Received signal for gnmi server to close")
-		s.Stop()
 		if serverControlValue == ServerStop {
+			s.ForceStop() // No graceful stop
 			stopSignalHandler <- true
 			log.Flush()
 			return
 		}
+		s.Stop() // Graceful stop
```
ganglyu, line 473: Why not stop signalHandler and flush log?

zbud-msft: s.Stop() is a graceful stop which is only reserved for cert rotation. Since we are not trying to exit process, just stop the server, we will not need to return or flush logs.

ganglyu (dismissed):

ganglyu (approved):

PR #228, opened 2024-05-07
Update pipeline to check memory leak

ganglyu:
#### Why I did it
We have found some memory leak introduced by cgo, and we need tools to detect memory leak.

#### How I did it
Run unit test with Address Santilizer to detect memory leak, and add a new step in pipeline to check memory leak.

#### How to verify it
Run sonic-gnmi pipeline.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -23,6 +23,8 @@ variables:
       value: $(System.PullRequest.TargetBranch)
     ${{ else }}:
       value: $(Build.SourceBranchName)
+  - name: UNIT_TEST_FLAG
+    value: 'ENABLE_TRANSLIB_WRITE=y'
 
 resources:
   repositories:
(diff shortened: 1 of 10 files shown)


liuh-80 (dismissed):

azure-pipelines.yml
```diff
@@ -151,8 +151,13 @@ stages:
 
     - script: |
         pushd sonic-gnmi
-        make check_gotest ENABLE_TRANSLIB_WRITE=y
-      displayName: "Test"
+        make all && make check_memleak ENABLE_TRANSLIB_WRITE=y
```
qiluo-msft, line 154: >all [](http://example.com/codeflow?start=13&length=3)

Do you need `ENABLE_TRANSLIB_WRITE` for `make all`? Since the flag is duplicated, can we define it in a central place?

ganglyu: I have updated the pipeline.

liuh-80 (approved):

PR #237, opened 2024-05-13
Improve GNMI incremental config and add unit test for BGP

ganglyu:
#### Why I did it
GCU has verified BGP configuration, we need to verify that GNMI can support the same BGP configuration.
Because GCU end2end test uses jsonpatch replace operation for BGP, we need to update GNMI replace operation for incremental config.
Golang does not have reliable library for jsonpatch.
Microsoft ADO: 27231737

#### How I did it
This unit test uses BGP configuration from GCU test.
This unit test generates GNMI request for BGP and use jsonpatch to verify patch file generated by GNMI server.
I also update gnmi replace implementation to support jsonpatch replace operation.
"config apply-patch" is equivalent to “config replace", so I use "config replace" for incremental config, and then we don't need to generate jsonpatch in golang.

#### How to verify it
Run gnmi unit test.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/sonic_data_client/mixed_db_client.go
+++ b/sonic_data_client/mixed_db_client.go
@@ -77,7 +77,6 @@ type MixedDbClient struct {
 	zmqAddress string
 	zmqClient swsscommon.ZmqClient
 	tableMap map[string]swsscommon.ProducerStateTable
-	zmqTableMap map[string]swsscommon.ZmqProducerStateTable
 	// swsscommon introduced dbkey to support multiple database
 	dbkey swsscommon.SonicDBKey
 	// Convert dbkey to string, namespace:container
@@ -97,8 +96,6 @@ var DbInstNum = 0
 
 func Hget(configDbConnector *swsscommon.ConfigDBConnector, table string, key string, field string) (string, error) {
     var fieldValuePairs = configDbConnector.Get_entry(table, key)
-    defer swsscommon.DeleteFieldValueMap(fieldValuePairs)
-
     if fieldValuePairs.Has_key(field) {
         return fieldValuePairs.Get(field), nil
     }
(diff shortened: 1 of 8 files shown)


test/test_gnmi_configdb.py
```diff
@@ -252,7 +252,7 @@ def test_gnmi_incremental_replace(self, test_data):
             test_path = item['path']
             test_value = item['value']
             for patch_data in patch_json:
-                assert patch_data['op'] == 'add', "Invalid operation"
```
qiluo-msft, line 255: >add [](http://example.com/codeflow?start=44&length=3)

you should also keep old testcase if it still make sense

sonic_data_client/mixed_db_client.go
```diff
@@ -1110,7 +1113,7 @@ func (c *MixedDbClient) handleTableData(tblPaths []tablePath) error {
 }
 
 /* Populate the JsonPatch corresponding each GNMI operation. */
-func (c *MixedDbClient) ConvertToJsonPatch(prefix *gnmipb.Path, path *gnmipb.Path, t *gnmipb.TypedValue, output *string) error {
+func (c *MixedDbClient) ConvertToJsonPatch(prefix *gnmipb.Path, path *gnmipb.Path, t *gnmipb.TypedValue, operation string, output *string) error {
```
qiluo-msft, line 1116: >operation [](http://example.com/codeflow?start=105&length=9)

Could you PR title or PR description for this new feature?

ganglyu: updated.

sonic_data_client/mixed_db_client.go
```diff
@@ -1122,11 +1125,7 @@ func (c *MixedDbClient) ConvertToJsonPatch(prefix *gnmipb.Path, path *gnmipb.Pat
 	}
 
 	elems := fullPath.GetElem()
-	if t == nil {
-		*output = `{"op": "remove", "path": "/`
-	} else {
-		*output = `{"op": "add", "path": "/`
-	}
+	*output = `{"op": "` + operation + `", "path": "/`
```
qiluo-msft, line 1128: >op [](http://example.com/codeflow?start=14&length=2)

Manual serialization is dangerous. Will you just construct the json object and use library to serialize it.

PR #299, opened 2024-10-11
Fix Wildcard entries to return empty data instead of invalid path

zbud-msft:

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


gnmi_server/server_test.go
```diff
@@ -3487,6 +3487,77 @@ func TestClientConnections(t *testing.T) {
 	}
 }
 
+func TestNonExistentTableNoError(t *testing.T) {
```
qiluo-msft, line 3490: >TestNonExistentTableNoError [](http://example.com/codeflow?start=5&length=27)

Is there a test case for existing table no error?

zbud-msft: Added test

qiluo-msft (approved):

xincunli-sonic (approved):

ganglyu (approved):

FengPan-Frank (approved):

gnmi_server/server_test.go
```diff
@@ -3487,6 +3487,169 @@ func TestClientConnections(t *testing.T) {
...
+
+			mutexNoti.Lock()
+
+			if len(gotNoti) == 0 {
+				t.Errorf("expected non zero notifications")
+			}
+
+			if diff := pretty.Compare(tt.wantNoti, gotNoti); diff != "" {
+				t.Log("\n Want: \n", tt.wantNoti)
+				t.Log("\n Got: \n", gotNoti)
```
qiluo-msft, line 3571: We also need to verify if the gnmi response is correct. 
Seems like a bug in your manual test: "{}"

zbud-msft: Added test to verify gnmi response is empty json, please take a look at EMPTY_JSON.txt.

It is not a bug, as gnmi_client as part of subscribe will print out json_ietf_value as json string and "{}" is a valid response.

json_ietf structured value should be a valid json string, which empty json represented as string, "{}", is. 
https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#23-structured-data-types

If we use gnmi_get which prints out json and not json string, we indeed get expected empty json
 ------------------------- 
 [elem {
  name: "PSU_INFO"
}
]
The GetResponse is below
-------------------------

{}
-------------------------

gechiang:
@qiluo-msft , @zbud-msft has made the changes per your review comment.  if the new changes looks good and no other comments, can you you help approve/merge this PR?
Thanks!

gechiang:
@BYGX-wcr , please help monitor this PR state.  When it is merged, please work with @zbud-msft to port this needed fix to internal 202205 chassis branch via patch.  Thanks!

qiluo-msft (approved):

PR #326, opened 2024-11-12
Update dbus timeout to 10 minutes

ganglyu:
#### Why I did it
We found that GCU takes about 10 minutes to update ACL rules

#### How I did it
Update dbus timeout to 10 minutes so that we can use GNMI to update ACL rules.

#### How to verify it
Run GNMI unit test and end to end test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


wen587 (approved):
do we need a global value so that everytime we just modify the single variable?

PR #332, opened 2024-12-19
Enable Pfcwd Queries

zbud-msft:
#### Why I did it

Pfcwd queries currently use PFC_WD_TABLE which is not the correct table name. Use PFC_WD instead.

Since Pfcwd are not crucial to other COUNTERS paths, if map creation of Pfcwd fails, we will fail gracefully and return empty json for Pfcwd instead of impacting all COUNTER_DB queries.

#### How I did it

Change table name

#### How to verify it

UT and manual testing

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/gnmi_server/server_test.go
+++ b/gnmi_server/server_test.go
@@ -242,6 +242,99 @@ func createKeepAliveServer(t *testing.T, port int64) *Server {
 	return s
 }
 
+func TestPFCWDErrors(t *testing.T) {
+	s := createServer(t, 8081)
+	go runServer(t, s)
+
+	mock := gomonkey.ApplyFunc(sdc.GetPfcwdMap, func() (map[string]map[string]string, error)  {
+		return nil, fmt.Errorf("Mock error")
+	})
+	defer mock.Reset()
+
+	fileName := "../testdata/COUNTERS:Ethernet_wildcard_alias.txt"
+	countersEthernetWildcardByte, err := ioutil.ReadFile(fileName)
+	if err != nil {
+		t.Fatalf("read file %v err: %v", fileName, err)
+	}
+	var countersEthernetWildcardJson interface{}
+	json.Unmarshal(countersEthernetWildcardByte, &countersEthernetWildcardJson)
+
+	tests := []struct {
+		desc    string
+		q       client.Query
+		wantNoti    []client.Notification
+		poll    int
+	}{
+		{
+			desc: "query COUNTERS/Ethernet*",
+			poll: 1,
+			q: client.Query{
+				Target: "COUNTERS_DB",
+				Type:    client.Poll,
+				Queries: []client.Path{{"COUNTERS", "Ethernet*"}},
+				TLS:     &tls.Config{InsecureSkipVerify: true},
+			},
+			wantNoti: []client.Notification{
+				client.Update{Path: []string{"COUNTERS", "Ethernet*"}, TS: time.Unix(0, 200), Val: countersEthernetWildcardJson},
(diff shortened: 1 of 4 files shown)


sonic_data_client/virtual_db.go
```diff
@@ -149,7 +149,10 @@ func getPfcwdMap() (map[string]map[string]string, error) {
 		}
 
 		for _, key := range resp {
-			name := key[13:]
+			if strings.Contains(key, "GLOBAL") || strings.Contains(key, "global") { // ignore PFC_WD|global / PFC_WD|GLOBAL
```
ganglyu, line 152: Why do we ignore PFC_WD|global?

zbud-msft: No counters associated with global since it is not a port. We are trying to create a mapping of oid from port name and pfc queue index which we will use to retrieve pfc wd counters. We ignore global since it serves no purpose for building the mapping since no port is associated with global

ganglyu (approved):

sonic_data_client/virtual_db.go
```diff
@@ -119,7 +119,7 @@ func initCountersPfcwdNameMap() error {
 }
 
 // Get the mapping between sonic interface name and oids of their PFC-WD enabled queues in COUNTERS_DB
-func getPfcwdMap() (map[string]map[string]string, error) {
+func GetPfcwdMap() (map[string]map[string]string, error) {
```
kperumalbfn, line 122: All other function names use camelcase format, should we go back to the previous name?

sonic_data_client/virtual_db.go
```diff
@@ -149,7 +149,10 @@ func getPfcwdMap() (map[string]map[string]string, error) {
 		}
 
 		for _, key := range resp {
-			name := key[13:]
+			if strings.Contains(key, "GLOBAL") || strings.Contains(key, "global") { // ignore PFC_WD|global / PFC_WD|GLOBAL
+				continue
+			}
+			name := key[7:]
```
kperumalbfn, line 155: instead of "7", should we get the length of "PFC_WD|" using a function for future proof?

gnmi_server/server_test.go
```diff
@@ -242,6 +242,106 @@ func createKeepAliveServer(t *testing.T, port int64) *Server {
...
+			c := client.New()
+			var gotNoti []client.Notification
+
+			q.NotificationHandler = func(n client.Notification) error {
+				mutexGotNoti.Lock()
+				if nn, ok := n.(client.Update); ok {
+					nn.TS = time.Unix(0, 200)
+					gotNoti = append(gotNoti, nn)
+				}
+				mutexGotNoti.Unlock()
```
kperumalbfn, line 302: connection close on error?

sonic_data_client/virtual_db.go
zbud-msft: Making this function public for testing. Capitalizing the first letter in the function name makes it public.

sonic_data_client/virtual_db.go
zbud-msft: Yes, I can make that change.

gnmi_server/server_test.go
zbud-msft: We have a defer statement that will stop the server connection, if there is an error, before the function exits, the defer statement will be called.

kperumalbfn (approved):

PR #335, opened 2025-01-07
Improve GNMI service to limit API access by role

ganglyu:
Microsoft ADO: 30073317

#### Why I did it
GNMI/GNOI services need to support role in API operation. RO user is not allowed to call write API.

#### How I did it
We have supported common name to role mapping, and I have updated API authentication to verify role.

#### How to verify it
Run unit test and end to end test.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/gnmi_server/server.go
+++ b/gnmi_server/server.go
@@ -91,6 +91,7 @@ type Config struct {
 
 var AuthLock sync.Mutex
 var maMu sync.Mutex
+const WriteAccessMode = "readwrite"
 
 func (i AuthTypes) String() string {
 	if i["none"] {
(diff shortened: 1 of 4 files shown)


gnmi_server/server.go
```diff
@@ -268,6 +269,13 @@ func authenticate(config *Config, ctx context.Context) (context.Context, error)
 		if err == nil {
 			success = true
 		}
+		// role must be readwrite to support write access
+		if writeAccess && len(rc.Auth.Roles) != 0 {
```
liuh-80, line 273: Check Roles length is not necessary, because it's already checked here:
https://github.com/sonic-net/sonic-gnmi/blob/c6f36b9bbe0360714df330f15fbdad0b88cec5a0/gnmi_server/clientCertAuth.go#L243

If you want only check roles for RW operation, you may also need check above code.

Suggest change to:

if writeAccess {
...
}

liuh-80 (dismissed):

liuh-80 (approved):

PR #351, opened 2025-01-31
GNOI Implementation of OS.Activate

hdwhdw:
#### Why I did it
Support OS.Activate for GNOI.

#### How I did it
Implement OS.Activate.

#### How to verify it
On physical switch:
```
admin@switch:~$ sudo sonic-installer list
Current: SONiC-OS-20240510.24
Next: SONiC-OS-master.762661-a3792f837
Available:
SONiC-OS-master.762661-a3792f837
SONiC-OS-20240510.24
admin@switch:~$ docker exec gnmi gnoi_client -target 127.0.0.1:50052 -logtostderr -insecure -module OS -rpc Activate -jsonin  '{"version": "SONiC-OS-20240510.24"}'
OS Activate
{"Response":{"ActivateOk":{}}}
admin@switch:~$ sudo sonic-installer list
Current: SONiC-OS-20240510.24
Next: SONiC-OS-20240510.24
Available:
SONiC-OS-master.762661-a3792f837
SONiC-OS-20240510.24
admin@switch:~$ docker exec gnmi gnoi_client -target 127.0.0.1:50052 -logtostderr -insecure -module OS -rpc Activate -jsonin  '{"version": "Foo"}'
OS Activate
{"Response":{"ActivateError":{"type":1,"detail":"Error: Image does not exist\n"}}}
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


gnmi_server/gnoi.go
```diff
@@ -157,6 +157,52 @@ func (srv *OSServer) Verify(ctx context.Context, req *gnoi_os_pb.VerifyRequest)
...
+	if err != nil {
+		log.Errorf("Failed to authenticate: %v", err)
+		return nil, err
+	}
+
+	log.Infof("gNOI: Activate")
+	image := req.GetVersion()
+	log.Infof("Requested to activate image %s", image)
+
+	dbus, err := ssc.NewDbusClient()
```
xincunli-sonic, line 171: >dbus [](http://example.com/codeflow?start=1&length=4)

Please call close after the client created?

defer dbus.Close()

gnmi_server/gnoi.go
```diff
@@ -157,6 +157,52 @@ func (srv *OSServer) Verify(ctx context.Context, req *gnoi_os_pb.VerifyRequest)
...
+		} else {
+			// Other error.
+			resp.Response = &gnoi_os_pb.ActivateResponse_ActivateError{
+				ActivateError: &gnoi_os_pb.ActivateError{
+					Type:    gnoi_os_pb.ActivateError_UNSPECIFIED,
+					Detail:  err.Error(),
+				},
+			}
+		}
+		return &resp, nil
```
xincunli-sonic, line 198: >nil [](http://example.com/codeflow?start=16&length=3)

Why would not we return actual error?

hdwhdw: I don't think this is a server error. See https://github.com/openconfig/gnoi/blob/main/os/os.proto#L372, this is an application level error instead of server error.

gnmi_server/gnoi.go
```diff
@@ -157,6 +157,52 @@ func (srv *OSServer) Verify(ctx context.Context, req *gnoi_os_pb.VerifyRequest)
...
+	log.Infof("Requested to activate image %s", image)
+
+	dbus, err := ssc.NewDbusClient()
+	if err != nil {
+		log.Errorf("Failed to create dbus client: %v", err)
+		return nil, err
+	}
+
+	var resp gnoi_os_pb.ActivateResponse
+	err = dbus.ActivateImage(image)
```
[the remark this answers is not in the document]
hdwhdw: I did try to do this previously but sonic-installer set-next-boot does not provide proper error code (it just return 1 instead, arguably it should return 2 (ENOENT)). The error code is propagated through DBUS and then to GNMI so I have to rely on the message. 

I will add a check here and correct this in both host-services and sonic-installer script.

gnmi_server/gnoi.go
hdwhdw: Done.

gnoi_client/gnoi_client.go
```diff
@@ -56,6 +56,8 @@ func main() {
 		switch *config.Rpc {
 		case "Verify":
 			gnoi_os.Verify(conn, ctx)
+		case "Activate":
```
qiluo-msft, line 59: >Activate [](http://example.com/codeflow?start=8&length=8)

wondering whether upstream repo already has a feature rich gnoi_client?

gnmi_server/gnoi.go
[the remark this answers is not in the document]
xincunli-sonic: I understand the sonic-installer set-next-boot command line just returned string, what I meant is actual place that call command line then it did wrap over there.

```
admin@vslab1:~$ sudo sonic-installer set-next-boot foo
Error: Image does not exist
```

gnmi_server/gnoi.go
xincunli-sonic: Okay, I saw the gnoi_client just handles response. Do you think is there any other caller?

gnmi_server/gnoi.go
```diff
@@ -161,6 +161,56 @@ func (srv *OSServer) Verify(ctx context.Context, req *gnoi_os_pb.VerifyRequest)
 	return resp, nil
 }
 
+func (srv *OSServer) Activate(ctx context.Context, req *gnoi_os_pb.ActivateRequest) (*gnoi_os_pb.ActivateResponse, error) {
+	_, err := authenticate(srv.config, ctx, false)
```
ganglyu, line 165: Need RW permission

gnmi_server/gnoi.go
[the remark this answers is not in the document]
hdwhdw: Doing it in https://github.com/sonic-net/sonic-host-services/pull/210, add a check for errno here. We can still keep the message check in an "either-or" condition just to be doubly safe.

gnmi_server/gnoi.go
hdwhdw: This function is a grpc internal server implementation, so the only caller is grpc server. If we return error, that means the grpc handler cannot proceed because we encountered an internal error. "Image not found" is not that and based on the server definition in openconfig, we should report that directly to the client, instead of making it a server failure. Whatever clients want to do with the response error is up to the client.

gnoi_client/gnoi_client.go
hdwhdw: Yes in https://google.github.io/gnxi/gnoi_os/.

gnmi_server/gnoi.go
hdwhdw: Done.

gnmi_server/gnoi.go
```diff
@@ -161,6 +161,56 @@ func (srv *OSServer) Verify(ctx context.Context, req *gnoi_os_pb.VerifyRequest)
...
+		return nil, err
+	}
+	defer dbus.Close()
+
+	var resp gnoi_os_pb.ActivateResponse
+	err = dbus.ActivateImage(image)
+	if err != nil {
+		log.Errorf("Failed to activate image %s: %v", image, err)
+		image_not_exists := os.IsNotExist(err) ||
+			(strings.Contains(strings.ToLower(err.Error()), "not") &&
```
FengPan-Frank, line 187: even with string check, could we use complete string match instead of "not" && "exist" which will have more corner cases.

hdwhdw: My main concern is error string equality check can easily have false negatives, say someone just remove a redundant space or any minor changes to the error string. Being that specific can make maintaining backend code difficult.

xincunli-sonic (approved):
LGTM

qiluo-msft (approved):

PR #360, opened 2025-02-13
Server side implementation of System.SetPackage

hdwhdw:
#### Why I did it
individual dpu upgrade https://github.com/sonic-net/SONiC/pull/1906

#### How I did it
implement server side of System.SetPackage

#### How to verify it
On physical switch:
```
admin@[REDACTED_HOST]:~$ sudo sonic-installer list
Current: SONiC-OS-20240510.24
Next: SONiC-OS-20240510.24
Available:
SONiC-OS-20240510.24
SONiC-OS-20231110.30
admin@[REDACTED_HOST]:~$ docker exec gnmi gnoi_client -target 127.0.0.1:50052 -logtostderr -insecure -module System -rpc SetPackage -package_filename=/tmp/sonic-mellanox.bin -package_version=foo -package_url=http://[REDACTED_IP]/sonic-mellanox.bin
System SetPackage

admin@[REDACTED_HOST]:~$ sudo sonic-installer list
Current: SONiC-OS-20240510.24
Next: SONiC-OS-20201231.25
Available:
SONiC-OS-20201231.25
SONiC-OS-20240510.24
admin@[REDACTED_HOST]:~$ ls /tmp/sonic-mellanox.bin
/tmp/sonic-mellanox.bin
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Server side implementation of System.SetPackage


gnmi_server/gnoi.go
```diff
@@ -337,15 +337,64 @@ func (srv *SystemServer) Traceroute(req *gnoi_system_pb.TracerouteRequest, rs gn
...
 	if err != nil {
 		return err
 	}
 	log.V(1).Info("gNOI: SetPackage")
-	return status.Errorf(codes.Unimplemented, "")
+
+	dbus, err := ssc.NewDbusClient()
+	if err != nil {
+		log.Errorf("Failed to create dbus client: %v", err)
+		return err
```
xincunli-sonic, line 352: >return err [](http://example.com/codeflow?start=2&length=10)

Suggestion: Return explicit gRPC status errors like status.Errorf(codes.Internal, "Error message...: %v", err) to give clearer context to gRPC clients.

gnmi_server/gnoi.go
```diff
@@ -337,15 +337,64 @@ func (srv *SystemServer) Traceroute(req *gnoi_system_pb.TracerouteRequest, rs gn
...
+		log.Errorf(errMsg)
+		return status.Errorf(codes.InvalidArgument, errMsg)
+	}
+
+	// Extract the package information
+	log.V(1).Infof("Received package information: %v", pkg.Package)
+
+	// Extract remote download information
+	download := pkg.Package.RemoteDownload
+	if download == nil {
```
xincunli-sonic, line 376: >download == nil [](http://example.com/codeflow?start=4&length=15)

 Also validate download.Path and pkg.Package.Filename are non-empty?

gnmi_server/gnoi.go
```diff
@@ -337,15 +337,64 @@ func (srv *SystemServer) Traceroute(req *gnoi_system_pb.TracerouteRequest, rs gn
...
+		return status.Errorf(codes.InvalidArgument, "RemoteDownload information is missing.")
+	}
+	// Download the package
+	err = dbus.DownloadImage(download.Path, pkg.Package.Filename)
+	if err != nil {
+		log.Errorf("Failed to download image: %v", err)
+		return err
+	}
+	// Install the package
+	err = dbus.InstallImage(pkg.Package.Filename)
```
ganglyu, line 387: How do we know this image is for DPU?

hdwhdw: Platform check is done by sonic_installer https://github.com/search?q=repo%3Asonic-net%2Fsonic-utilities%20verify_image_platform&type=code

ganglyu: It's possible to use SetPackage to install different firmware, as sonic-installer does not support other firmware types. Therefore, it is necessary to identify the image type.

ganglyu: If active is false, we should download but should not install

gnmi_server/gnoi.go
hdwhdw: Done.

gnmi_server/gnoi.go
hdwhdw: Interesting. The initial design is SetPackage should carry at least the effect of "sonic-installer install", even if "activate=false", and if "activate=true", then it should also "set-next-boot". 

I feel like your proposal is a bit granular, basically if remote_download is provided, we download, and if activate=true, we install (which will also set-next-boot). Updated.

hdwhdw: But this will require the generic/dpu upgrade process to be updated with either activate=true, or two separate SetPackage calls, one for download and one for install.

gnmi_server/gnoi.go
hdwhdw: Done.

gnmi_server/gnoi_system.go
```diff
@@ -298,15 +298,93 @@ func (srv *Server) Traceroute(req *syspb.TracerouteRequest, stream syspb.System_
...
+	// Validate request type
+	pkg, ok := req.GetRequest().(*syspb.SetPackageRequest_Package)
+	if !ok {
+		errMsg := fmt.Sprintf("invalid request type: %T, expected SetPackageRequest_Package", req.GetRequest())
+		log.Errorf(errMsg)
+		return status.Errorf(codes.InvalidArgument, errMsg)
+	}
+
+	// A filename and a version must be provided
+	if pkg.Package.Filename == "" {
```
qiluo-msft, line 335: >Filename [](http://example.com/codeflow?start=16&length=8)

Will you add code to check `Package.remote_download`? I understand we only want to implement remote download.

gnmi_server/gnoi_system.go
```diff
@@ -298,15 +298,93 @@ func (srv *Server) Traceroute(req *syspb.TracerouteRequest, stream syspb.System_
...
+	log.V(1).Infof("Package filename: %s, version: %s", pkg.Package.Filename, pkg.Package.Version)
+
+	// Download the package if RemoteDownload is provided
+	if pkg.Package.RemoteDownload != nil {
+		// Validate RemoteDownload
+		log.V(1).Infof("RemoteDownload provided")
+		// Check if the path is provided
+		if pkg.Package.RemoteDownload.Path == "" {
+			log.Errorf("RemoteDownload path is missing")
+			return status.Errorf(codes.InvalidArgument, "remote download path is missing")
```
qiluo-msft, line 353: >InvalidArgument [](http://example.com/codeflow?start=30&length=15)

by protocol, is "Path" always required? If not, we may return NotImplemented error.

gnmi_server/gnoi_system.go
```diff
@@ -298,15 +298,93 @@ func (srv *Server) Traceroute(req *syspb.TracerouteRequest, stream syspb.System_
...
+		// Download the package
+		err = dbus.DownloadImage(pkg.Package.RemoteDownload.Path, pkg.Package.Filename)
+		if err != nil {
+			log.Errorf("Failed to download image: %v", err)
+			return status.Errorf(codes.Internal, "failed to download image: %v", err)
+		}
+		log.V(1).Infof("Package %s downloaded successfully to %s", pkg.Package.Version, pkg.Package.Filename)
+	}
+
+	// If activate is requested, install the package and set it to be the next boot image
```
qiluo-msft, line 366: >If [](http://example.com/codeflow?start=4&length=2)

If not, what is the expectation? Install the new image but not set as default boot image?

hdwhdw: The proto service [definition](https://github.com/openconfig/gnoi/blob/main/system/system.proto#L328) is somewhat ambiguous here. So what we do is we just download the image and put it into where "filename" specifies (notice that in the proto definition filename is a full path), without installing or setting it as next boot image. Currently not all bootloader can separate installing and setting it as next boot image so for now this is the best thing we can do.

gnmi_server/gnoi_system.go
hdwhdw: This is always required as the service definition is 
```
// SetPackageRequest will place the package onto the target and optionally mark
// it as the next bootable image. The initial message must be a package
// message containing the filename and information about the file.
```

gnmi_server/gnoi_system.go
hdwhdw: Just to clarify, here if you enable "activate", it will install it and set it to default.

qiluo-msft: It is arguable whether we should "install without active" during download step, or during activate step. Is there any trade-off?

hdwhdw: We should at least "place the package onto the target", which means at least download regardless `activate = true` or not. And optionally if `activate = true`, we should complete the remaining steps, i.e. install and set-next-boot.

hdwhdw: The current problem is that we don't have a support for "install without activate",  (if there is, I will do it at the download steps". We can support this but it will requrie more works in installation util and possibly bootloader.

hdwhdw: The trade off i can think of:
* installation at download:
   1. if installation fails, we can clean up and save disk.
   2. but you cannot rollback from the client (since system state has changed)
 * installation at activation:
   1. easier to rollback the "Setpackage without install step"
   2. can  SetPackage multiple sonic image on the host and optionally activate one of them.

hdwhdw: So in general I think installation at download is a bit safer because you can catch error and avoid having a bad image occupying hard disk.

gnmi_server/gnoi_system.go
```diff
@@ -298,15 +298,93 @@ func (srv *Server) Traceroute(req *syspb.TracerouteRequest, stream syspb.System_
...
+		log.V(1).Infof("RemoteDownload provided")
+		// Check if the path is provided
+		if pkg.Package.RemoteDownload.Path == "" {
+			log.Errorf("RemoteDownload path is missing")
+			return status.Errorf(codes.InvalidArgument, "remote download path is missing")
+		}
+		log.V(1).Infof("RemoteDownload path: %s", pkg.Package.RemoteDownload.Path)
+
+		// Download the package
+		err = dbus.DownloadImage(pkg.Package.RemoteDownload.Path, pkg.Package.Filename)
```
ganglyu, line 358: Do you download the image to this specified filename? Is it possible for someone to use this API to attack the SONiC system?

gnmi_server/gnoi_system.go
```diff
@@ -298,15 +298,93 @@ func (srv *Server) Traceroute(req *syspb.TracerouteRequest, stream syspb.System_
...
+			return status.Errorf(codes.Internal, "failed to install image: %v", err)
+		}
+		log.V(1).Infof("Package %s installed successfully", pkg.Package.Filename)
+		// Currently, Installing the image will automatically set it as the next boot image
+		log.V(1).Infof("Package %s set as next boot image", pkg.Package.Filename)
+	}
+
+	// Send response to client
+	if err := rs.SendAndClose(&syspb.SetPackageResponse{}); err != nil {
+		log.Errorf("Failed to send response: %v", err)
```
FengPan-Frank, line 382: why not return grpc error for this?

gnmi_server/gnoi_system.go
hdwhdw: the gnoi server doesn't verify, but the host service will permit the download only if the directory is all writeable:
https://github.com/sonic-net/sonic-host-services/blob/master/host_modules/image_service.py#L57 which should make the sonic system safe.

gnmi_server/gnoi_system.go
hdwhdw: This err is returned by grpc framework so it is the grpc code.

gnmi_server/gnoi_system.go
ganglyu: But since sonic-host-services is running as root, it should have permission to write to all directories.

zbud-msft: If we use external gnoi client to call SetPackage to install malicious image, root permissions are not needed?

hdwhdw: Thanks for pointing out. now only allow downloading into /lib/firmware/

gnmi_server/gnoi_system_test.go
```diff
@@ -4,20 +4,427 @@ import (
...
+	mockServer.EXPECT().Context().Return(ctx).AnyTimes()
+
+	t.Run("SetPackageSuccessDownloadOnly", func(t *testing.T) {
+		mockServer.EXPECT().Recv().Return(&syspb.SetPackageRequest{
+			Request: &syspb.SetPackageRequest_Package{
+				Package: &syspb.Package{
+					RemoteDownload: &commonpb.RemoteDownload{
+						Path: "http://example.com/package",
+					},
+					Filename: "/lib/firmware/package.bin",
```
qiluo-msft, line 51: >/lib/firmware/ [](http://example.com/codeflow?start=16&length=14)

This path is not discussed in HLD. Let's finalize the design and update HLD first.

gnmi_server/gnoi_system.go
hdwhdw: @ganglyu actually the root user only *bypass* the permission check, not having a different permission, the sonic-host-service code still would return false to sensitive location such as /etc/sonic. (there is a single permission mask for a file indicating owner, group and others access, so whether you are root or someone else don't matter).

gnmi_server/gnoi_system_test.go
hdwhdw: Reverted because the basis of the /lib/firmware/ change (to prevent attacking sonic system) is moot, the permission check in sonic-host-services (must write to a directory that is all writable) is sufficient to prevent writing to places like /etc/sonic

ganglyu (approved):

Issue #357, opened 2025-02-13
sonic-clients: Allow for ConfigReplace to run after full config push over gNMI without a device reboot/reload


anders-nexthop:
Today, there is a gNMI endpoint available that allows the operator to push a full config remotely: `/sonic-db:CONFIG_DB/localhost/`

However, when you send a file to this path, it just stores the configuration in the file system without making any service changes. A subsequent reboot/reload is required for these changes to take effect.
 
GCU has the ability to diff a config file against the current running config, apply only necessary modifications to the running config, and then restart or update only the specific set of services that are affected by the change. This can be done with the `config replace` cli command, or programmatically using [ConfigReplacer](https://github.com/sonic-net/sonic-utilities/blob/7f3957c43664082ee85d0e4b334d6b838ef861a4/generic_config_updater/generic_updater.py#L159) class from sonic-utilities/generic_config_updater/generic_updater.py.

The `config replace` command is available over D-Bus, along with `config reload`:

```
$ sudo busctl introspect org.SONiC.HostService.config /org/SONiC/HostService/config
NAME                                TYPE      SIGNATURE RESULT/VALUE FLAGS
org.SONiC.HostService.config        interface -         -            -
.reload                             method    s         is           -
.save                               method    s         is           -
org.freedesktop.DBus.Introspectable interface -         -            -
.Introspect                         method    -         s            -
$ sudo busctl introspect org.SONiC.HostService.gcu /org/SONiC/HostService/gcu
NAME                                TYPE      SIGNATURE RESULT/VALUE FLAGS
org.SONiC.HostService.gcu           interface -         -            -
.apply_patch_db                     method    s         is           -
.apply_patch_yang                   method    s         is           -
.create_checkpoint                  method    s         is           -
.delete_checkpoint                  method    s         is           -
.replace_db                         method    s         is           -
.replace_yang                       method    s         is           -
org.freedesktop.DBus.Introspectable interface -         -            -
.Introspect                         method    -         s            -
```
 The `.replace_db` D-Bus endpoint will trigger `config replace`. The gNMI service client already uses several of these D-Bus endpoints, so it would be straightforward to add a new handler to [sonic-gnmi/sonic_service_client/dbus_client.go](https://github.com/sonic-net/sonic-gnmi/blob/master/sonic_service_client/dbus_client.go) that triggers `config replace`. 

[...]

PR #369, opened 2025-03-19
Issue #357: Use ConfigReplace when full config is sent over gNMI

anders-nexthop:
#### Why I did it

By default, gNMI does a version of `config reload` when passed a full config. This is disruptive, since it requires services to be restarted for it to take effect. It's also not intuitive; all of the other SET gnmi commands take place immediately. By enhancing the full config endpoint handling to trigger `config replace` instead of the gnmi-specific version of reload that exists today, we resolve both of those issues.

The existing implementation uses a full config DELETE given alongside a full config UPDATE to represent a reload. We can leave that syntax and behavior intact and instead add logic to handle a full config REPLACE.

fixes [Issue 357](https://github.com/sonic-net/sonic-gnmi/issues/357)

#### How I did it

There is already a D-BUS endpoint implemented for `config replace`, so all this takes is a handler implemented in [gnmi/sonic_service_client/dbus_client.go](https://github.com/sonic-net/sonic-gnmi/blob/master/sonic_service_client/dbus_client.go) which will pass the full config it receives through to HostServices (and from there to GCU).

#### How to verify it

Manually verified by sending full configs using `gnmic` and ensuring that state changes are propagated to ConfigDb.

DBUS message passing verified by using `dbus-monitor` to verify content of messages being passed.

All sonic-gnmi unit tests pass. New test cases were added to verify the SET-REPLACE handler logic. A new unit test was added to `dbus_client_test.go` to test the new D-Bus endpoint.

A sonic-mgmt testcase for this behavior is proposed [here](https://github.com/sonic-net/sonic-mgmt/issues/17885).

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Use ConfigReplace when full config is sent over gNMI

#### A picture of a cute animal (not mandatory but encouraged)

   _
 / . . \                        
  \   /.   \__________/ \
    |    \--------------/
    ^

(be kind i drew it myself)

### DIFF
--- a/sonic_data_client/mixed_db_client.go
+++ b/sonic_data_client/mixed_db_client.go
@@ -1400,6 +1400,28 @@ func (c *MixedDbClient) SetFullConfig(delete []*gnmipb.Path, replace []*gnmipb.U
 	return nil
 }
 
+func (c *MixedDbClient) ReplaceFullConfig(delete []*gnmipb.Path, replace []*gnmipb.Update, update []*gnmipb.Update) error {
+	val := replace[0].GetVal()
+	ietf_json_val := val.GetJsonIetfVal()
+	if len(ietf_json_val) == 0 {
+		return fmt.Errorf("Value encoding is not IETF JSON")
+	}
+	content := []byte(ietf_json_val)
+
+	var sc ssc.Service
+	sc, err := ssc.NewDbusClient()
+	if err != nil {
+		return err
+	}
+
+	err = sc.ConfigReplace(string(content))
+
+	if err == nil {
+		err = sc.ConfigSave("/etc/sonic/config_db.json")
+	}
+	return err
+}
+
 func (c *MixedDbClient) SetDB(delete []*gnmipb.Path, replace []*gnmipb.Update, update []*gnmipb.Update) error {
 	/* DELETE */
 	deletePathList, err := c.getAllDbtablePath(delete)
@@ -1459,6 +1481,14 @@ func (c *MixedDbClient) SetConfigDB(delete []*gnmipb.Path, replace []*gnmipb.Upd
 		}
 		if (len(deletePath.GetElem()) == 0) && (len(updatePath.GetElem()) == 0) {
 			return c.SetFullConfig(delete, replace, update)
+		}
+	} else if deleteLen == 0 && replaceLen == 1 && updateLen == 0 {
+		replacePath, err := c.gnmiFullPath(c.prefix, replace[0].GetPath())
+		if err != nil {
+			return err
+		}
+		if len(replacePath.GetElem()) == 0 {
+			return c.ReplaceFullConfig(delete, replace, update)
 		}
 	}
 	return c.SetIncrementalConfig(delete, replace, update)
--- a/common_utils/context.go
+++ b/common_utils/context.go
@@ -58,6 +58,7 @@ const (
 	DBUS_IMAGE_INSTALL
 	DBUS_IMAGE_LIST
 	DBUS_IMAGE_ACTIVATE
+	DBUS_CONFIG_REPLACE
 	COUNTER_SIZE
 )
 
@@ -103,6 +104,8 @@ func (c CounterType) String() string {
 		return "DBUS image list"
 	case DBUS_IMAGE_ACTIVATE:
 		return "DBUS image activate"
+	case DBUS_CONFIG_REPLACE:
+		return "DBUS config replace"
 	default:
 		return ""
 	}
--- a/sonic_service_client/dbus_client.go
+++ b/sonic_service_client/dbus_client.go
@@ -16,6 +16,7 @@ type Service interface {
 
 	// SONiC Host Service D-Bus API
 	ConfigReload(fileName string) error
+	ConfigReplace(fileName string) error
 	ConfigSave(fileName string) error
 	ApplyPatchYang(fileName string) error
 	ApplyPatchDb(fileName string) error
@@ -124,6 +125,16 @@ func (c *DbusClient) ConfigReload(config string) error {
 	busPath := c.busPathPrefix + modName
 	intName := c.intNamePrefix + modName + ".reload"
 	_, err := DbusApi(busName, busPath, intName, 60, config)
+	return err
+}
+
+func (c *DbusClient) ConfigReplace(config string) error {
+	common_utils.IncCounter(common_utils.DBUS_CONFIG_REPLACE)
+	modName := "gcu"
+	busName := c.busNamePrefix + modName
+	busPath := c.busPathPrefix + modName
+	intName := c.intNamePrefix + modName + ".replace_db"
+	_, err := DbusApi(busName, busPath, intName, 600, config)
 	return err
 }
 
--- a/sonic_service_client/dbus_client_test.go
+++ b/sonic_service_client/dbus_client_test.go
@@ -223,6 +223,67 @@ func TestConfigReloadTimeout(t *testing.T) {
 	}
 }
 
+func TestConfigReplace(t *testing.T) {
+	mock1 := gomonkey.ApplyFunc(dbus.SystemBus, func() (conn *dbus.Conn, err error) {
+		return &dbus.Conn{}, nil
+	})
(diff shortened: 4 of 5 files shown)


sonic_data_client/mixed_db_client.go
```diff
@@ -1385,19 +1385,23 @@ func (c *MixedDbClient) SetFullConfig(delete []*gnmipb.Path, replace []*gnmipb.U
 		return fmt.Errorf("Value encoding is not IETF JSON")
 	}
 	content := []byte(ietf_json_val)
+
+	// still needed...?
```
qiluo-msft, line 1389: >// still needed...? [](http://example.com/codeflow?start=1&length=19)

This is a good question. I guess ConfigReplace already validating running config?

anders-nexthop: Yes, ConfigReplace does that validation. It also ensures atomicity, so no config changes will be made unless the whole config is valid.

I left this in to preserve the existing behavior, at least to some extent. So on a gNOI Reboot, the config passed here will be picked up. But I'm not sure if we want to keep that behavior or not.

ganglyu: We can remove the yang validation

ganglyu:
@anders-nexthop 
https://github.com/sonic-net/sonic-gnmi/blob/master/sonic_data_client/mixed_db_client.go#L1454-L1463
For now, we use special syntax for a full configuration update. When we switch to using config replace, we should also change the syntax to replace "/".

anders-nexthop:
> @anders-nexthop https://github.com/sonic-net/sonic-gnmi/blob/master/sonic_data_client/mixed_db_client.go#L1454-L1463 For now, we use special syntax for a full configuration update. When we switch to using config replace, we should also change the syntax to replace "/".

I don't quite follow. I thought a full config update was represented with both and empty 'delete' path and an empty 'update' path. Is that not correct? What would we do differently to represent a config replace? If we can represent this operation in a different way we can leave the existing implementation in place, which may be desirable for compatibility. Are you perhaps suggesting that we could instead use the 'replace' path, passing in an empty path with the file contents as a value? For example:
```
    # from sonic-mgmt/tests/gnmi/test_gnmi_configdb.py
    delete_list = ["/sonic-db:CONFIG_DB/localhost/"]
    update_list = ["/sonic-db:CONFIG_DB/localhost/:@/root/%s" % filename]
    replace_list = ["/sonic-db:CONFIG_DB/localhost/:@/root/%s" % filename]

    gnmi_set(duthost, ptfhost, delete_list, update_list, [])                                       # this is the existing syntax

    gnmi_set(duthost, ptfhost, [], [], replace_list)                                                     # this would trigger ConfigReplace
```

ganglyu:
@anders-nexthop 
yes, you are correct, we could instead use the 'replace' path, passing in an empty path with the file contents as a value

anders-nexthop:
> @anders-nexthop yes, you are correct, we could instead use the 'replace' path, passing in an empty path with the file contents as a value

Yes that makes sense. I will make that change and update the PR.

sonic_data_client/mixed_db_client.go
anders-nexthop: I will remove the temp file.

I also see that in SetIncrementalConfig() above, we call ConfigSave() if the configuration is successfully accepted, so that `/etc/sonic/config_db.json` is updated. Should we do that here as well

ganglyu:
yes, this operation should be persistent

sonic_data_client/mixed_db_client.go
anders-nexthop: I added a call to ConfigSave() after the ConfigReload() completes successfully to make this command persistent.

qiluo-msft:
> > > @anders-nexthop https://github.com/sonic-net/sonic-gnmi/blob/master/sonic_data_client/mixed_db_client.go#L1454-L1463 For now, we use special syntax for a full configuration update. When we switch to using config replace, we should also change the syntax to replace "/".
> > 
> > 
> > I don't quite follow. I thought a full config update was represented with both and empty 'delete' path and an empty 'update' path. Is that not correct? What would we do differently to represent a config replace? If we can represent this operation in a different way we can leave the existing implementation in place, which may be desirable for compatibility. Are you perhaps suggesting that we could instead use the 'replace' path, passing in an empty path with the file contents as a value? For example:
> > ```
> >     # from sonic-mgmt/tests/gnmi/test_gnmi_configdb.py
> >     delete_list = ["/sonic-db:CONFIG_DB/localhost/"]
> >     update_list = ["/sonic-db:CONFIG_DB/localhost/:@/root/%s" % filename]
> >     replace_list = ["/sonic-db:CONFIG_DB/localhost/:@/root/%s" % filename]
> > 
> >     gnmi_set(duthost, ptfhost, delete_list, update_list, [])                                       # this is the existing syntax
> > 
> >     gnmi_set(duthost, ptfhost, [], [], replace_list)                                                     # this would trigger ConfigReplace
> > ```
> 
> @ganglyu @qiluo-msft I have updated the PR to be based off of the `replace_list`, which keeps the original behavior intact if using the the delete_list/update_list syntax. The build is failing with an issue that seems unrelated:
> 
> ```
> error: sonic-interface:29: module "sonic-vnet" not found in search path
> ```
> 
> I added new test cases to cover this change. They are passing locally but do to the build problem referenced above they have not been run through the pipeline yet.

Seems like the build failure is resolved by https://github.com/sonic-net/sonic-buildimage/pull/22293

anders-nexthop:
> > > > @anders-nexthop https://github.com/sonic-net/sonic-gnmi/blob/master/sonic_data_client/mixed_db_client.go#L1454-L1463 For now, we use special syntax for a full configuration update. When we switch to using config replace, we should also change the syntax to replace "/".
> > > 
> > > 
> > > I don't quite follow. I thought a full config update was represented with both and empty 'delete' path and an empty 'update' path. Is that not correct? What would we do differently to represent a config replace? If we can represent this operation in a different way we can leave the existing implementation in place, which may be desirable for compatibility. Are you perhaps suggesting that we could instead use the 'replace' path, passing in an empty path with the file contents as a value? For example:
> > > ```
> > >     # from sonic-mgmt/tests/gnmi/test_gnmi_configdb.py
> > >     delete_list = ["/sonic-db:CONFIG_DB/localhost/"]
> > >     update_list = ["/sonic-db:CONFIG_DB/localhost/:@/root/%s" % filename]
> > >     replace_list = ["/sonic-db:CONFIG_DB/localhost/:@/root/%s" % filename]
> > > 
> > >     gnmi_set(duthost, ptfhost, delete_list, update_list, [])                                       # this is the existing syntax
> > > 
> > >     gnmi_set(duthost, ptfhost, [], [], replace_list)                                                     # this would trigger ConfigReplace
> > > ```
> > 
> > 
> > @ganglyu @qiluo-msft I have updated the PR to be based off of the `replace_list`, which keeps the original behavior intact if using the the delete_list/update_list syntax. The build is failing with an issue that seems unrelated:
> > ```
> > error: sonic-interface:29: module "sonic-vnet" not found in search path
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I added new test cases to cover this change. They are passing locally but do to the build problem referenced above they have not been run through the pipeline yet.
> 
> Seems like the build failure is resolved by [sonic-net/sonic-buildimage#22293](https://github.com/sonic-net/sonic-buildimage/pull/22293)

@qiluo-msft the build failed again with a similar failure:

[error: /usr/local/cvlyang-models/sonic-vnet.yang:18: module "sonic-vxlan" not found in search path
](https://dev.azure.com/mssonic/build/_build/results?buildId=823423&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=335d3395-6d29-5762-4daa-c188603f3eb8&l=267)

anders-nexthop:
@qiluo-msft and @ganglyu I believe I have responded to all of the outstanding review comments, can I get approval for this change?

qiluo-msft (approved):

ganglyu (approved):

PR #371, opened 2025-04-04
Manually revert delete field logic

zbud-msft:
#### Why I did it

#31673482

Race condition with deleting field from map while another thread reads from the same map. Causes fatal error concurrent map read and write. Reverting this logic for now and will reimplement with better synchronization logic.

#### How I did it

Manually revert since auto revert doesn't work.

#### How to verify it

Manual test.

202411

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


qiluo-msft (approved):

qiluo-msft:
Please update "How to verify it" in PR description.

zbud-msft:
Delete logic that causes race conditon deleted in #375

PR #375, opened 2025-04-15
Fix poll mode for querying non existing table or key for APPL_DB

zbud-msft:
#### Why I did it

SONiC telemetry does not behave according to gnmi specifications

| **POLL MODE**                   | **Target doesn't exist**                                      | **Table doesn't exist**                                                                                                              | **Key doesn't exist**                                                                                                      | **Table gets deleted**                        | **Key gets deleted**                                                                               | **Field added**                | **Field gets changed**            |
|---------------------------------|----------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------|---------------------------------|-------------------------------------|
| **Official gNMI Spec**          | RPC_NOT_FOUND, entire RPC fails                               | Omit Notification from RPC response, rpc should be open                                                                              | Omit notification from RPC response, rpc should be open                                                                     | Delete notifcation includes path               | Delete notification includes path                                                                                             | Normal update response         | Normal update response             |
| **SONiC**                       | RPC_NOT_FOUND, entire RPC fails                               | RPC stays open and updates work properly. We do send empty notifications which isn't technically part of the official gnmi protocol. | Server fails internally with redis Hget, client is left to hang with no sync response; RPC is left in a bad state with no updates even from good paths. | Empty update notifications. No delete proto    | Update notification with remaining keys or just empty update notification. No delete proto                                     | Normal update response         | Normal Update response             |

https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#3513-the-subscription-message

As per gnmi specification regarding missing data, "There is no requirement that the path specified in the message must exist within the current data tree on the server. While the path within the subscription SHOULD be a valid path within the set of schema modules that the target supports, subscribing to any syntactically valid path within such modules MUST be allowed. In the case that a particular path does not (yet) exist, the target MUST NOT close the RPC and instead should continue to monitor for the existence of the path, and transmit telemetry updates should it exist in the future."

As per gnmi specification regarding deleted keys/paths, "Where a node within the subscribed paths has been removed, the delete field of the Notification message MUST have the path of the node that has been removed appended to it."

#### How I did it

If there is missing data, we will not return any failures, but instead we will send sync responses only until data is on the device for polling APPL_DB

If data is deleted, we will send a delete notification to tell client that path they are querying is deleted and after send sync responses for polling APPL_DB.

#### How to verify it

UT

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


sonic_data_client/db_client.go
```diff
@@ -300,37 +300,59 @@ func (c *DbClient) PollRun(q *queue.PriorityQueue, poll chan struct{}, w *sync.W
 	c.q = q
 	c.channel = poll
 
+	prevUpdates := make(map[string]bool)
```
qiluo-msft, line 303: >prevUpdates [](http://example.com/codeflow?start=1&length=11)

This map will always increase size in lifetime of telemetry process. Why you need it?

qiluo-msft:
> func (c *DbClient) PollRun(q *queue.PriorityQueue, poll chan struct{}, w *sync.WaitGroup, subscribe *gnmipb.SubscriptionList) {

Could you add testcases which behavior different on old and new code?

---
Refers to: sonic_data_client/db_client.go:297 in 6e6c3ca. [](commit_id = 6e6c3cac7b2ce54f506a59817f92635a53189981, deletion_comment = False)

sonic_data_client/db_client.go
```diff
@@ -863,7 +888,7 @@ func Msi2TypedValue(msi map[string]interface{}) (*gnmipb.TypedValue, error) {
 		}}, nil
 }
 
-func tableData2TypedValue(tblPaths []tablePath, op *string) (*gnmipb.TypedValue, error) {
+func tableData2TypedValue(tblPaths []tablePath, op *string) (*gnmipb.TypedValue, error, bool) {
```
hdwhdw, line 882: Can you leave some documentation on what exactly is this bool? Reader might want to check whether the logic around this bool is sound but it is not possible here.

sonic_data_client/db_client.go
```diff
@@ -884,23 +909,28 @@ func tableData2TypedValue(tblPaths []tablePath, op *string) (*gnmipb.TypedValue,
...
 			}
 		}
 		err := TableData2Msi(&tblPath, useKey, nil, &msi)
+
 		if err != nil {
-			return nil, err
+			return nil, err, false
+		}
+		if len(msi) == 0 {
+			return nil, nil, true
```
hdwhdw, line 920: Question: Why do we want to return early here?

hdwhdw (commented):
Can you explain the background of the change. It looks like to me that we are adding entries whenever someone poll an nonexistent table or key. Is this okay? Shouldn't failure be the default option here?

zbud-msft:
> Can you explain the background of the change. It looks like to me that we are adding entries whenever someone poll an nonexistent table or key. Is this okay? Shouldn't failure be the default option here?

Updated the PR description with more details. We are sending sync responses only to keep the RPC available and to let the client know that there is no data available but we will send data when it becomes available. Failure and closing RCP connection (current behavior) does not conform to official gnmi spec

sonic_data_client/db_client.go
hdwhdw: Thanks. Nits: negative predicate is in general discouraged because it creates confusion for readers and is error prone. i.e. API like hasData() is much better than NoData().  Consider negating this definition.

sonic_data_client/db_client.go
hdwhdw: Thanks!

sonic_data_client/db_client.go
zbud-msft: Agree, will change.

sonic_data_client/non_db_client.go
```diff
@@ -553,6 +553,11 @@ func (c *NonDbClient) PollRun(q *queue.PriorityQueue, poll chan struct{}, w *syn
 		log.V(4).Infof("Sync done, poll time taken: %v ms", int64(time.Since(t1)/time.Millisecond))
 	}
 }
+
+func (c *NonDbClient) AppDBPollRun(q *queue.PriorityQueue, poll chan struct{}, w *sync.WaitGroup, subscribe *gnmipb.SubscriptionList) {
```
qiluo-msft, line 557: > { [](http://example.com/codeflow?start=133&length=2)

Please fix the indentation.

sonic_data_client/db_client.go
```diff
@@ -39,6 +39,8 @@ type Client interface {
 	// The service will stop upon detection of poll channel closing.
 	// It should run as a go routine
 	PollRun(q *queue.PriorityQueue, poll chan struct{}, w *sync.WaitGroup, subscribe *gnmipb.SubscriptionList)
+	// Poll to service AppDB only
+	AppDBPollRun(q *queue.PriorityQueue, poll chan struct{}, w *sync.WaitGroup, subscribe *gnmipb.SubscriptionList)
```
qiluo-msft, line 43: >AppDBPollRun [](http://example.com/codeflow?start=1&length=12)

I could not find the code calling it, do you miss some file?

zbud-msft: Fixed

sonic_data_client/db_client.go
zbud-msft: Fixed.

qiluo-msft:
LGTM. Please test on both single-asic, T2 multi-asic DUTs.

gnmi_server/client_subscribe.go
```diff
@@ -195,7 +195,11 @@ func (c *Client) Run(stream gnmipb.GNMI_SubscribeServer) (err error) {
 		c.polled = make(chan struct{}, 1)
 		c.polled <- struct{}{}
 		c.w.Add(1)
-		go dc.PollRun(c.q, c.polled, &c.w, c.subscribe)
+		if strings.Contains(target, "APPL_DB") {
```
qiluo-msft, line 198: >strings.Contains [](http://example.com/codeflow?start=5&length=16)

too relaxing condition, please be more specific, check the single-asic or multi-asic APPL_DB only.

zbud-msft: Done

qiluo-msft (approved):

gnmi_server/client_subscribe.go
```diff
@@ -195,7 +195,11 @@ func (c *Client) Run(stream gnmipb.GNMI_SubscribeServer) (err error) {
 		c.polled = make(chan struct{}, 1)
 		c.polled <- struct{}{}
 		c.w.Add(1)
-		go dc.PollRun(c.q, c.polled, &c.w, c.subscribe)
+		if target == "APPL_DB" || strings.HasPrefix(target, "APPL_DB/") {
```
FengPan-Frank, line 198: is this only applicable for APPL_DB?

sonic_data_client/db_client.go
```diff
@@ -294,6 +296,69 @@ func streamSampleSubscription(c *DbClient, sub *gnmipb.Subscription, updateOnly
...
+		if !more {
+			log.V(1).Infof("%v poll channel closed, exiting pollDb routine", c)
+			return
+		}
+		t1 := time.Now()
+
+		for gnmiPath, tblPaths := range c.pathG2S {
+			pathKey := fmt.Sprintf("%v", gnmiPath)
+			val, err, updateReceived := AppDBTableData2TypedValue(tblPaths, nil)
+			if !updateReceived { // No updates sent for missing data
```
hdwhdw, line 318: Is this comment stale?

hdwhdw (approved):

gnmi_server/client_subscribe.go
zbud-msft: For now just APP DB, will integrate these changes for all DB as a TODO

sonic_data_client/db_client.go
zbud-msft: No, I changed from NoUpdate (where I used negation wording) to !UpdateReceived, since we now return true if we have data versus true if we have no data.

gnmi_server/client_subscribe.go
zbud-msft: Issue: https://github.com/sonic-net/sonic-gnmi/issues/379

sonic_data_client/db_client.go
hdwhdw: Ah thanks!

r12f:
merged in 202411. fixing tag.

PR #139, opened 2023-07-21
Add delete field to On change response when key is deleted

zbud-msft:
#### Why I did it

Currently when node is deleted, we do not send a delete field which should be the case according to gnmi specification:

https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#21-reusable-notification-message-format

#### How I did it

Added delete field

#### How to verify it

UT

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

PR #383, opened 2025-04-23
Remove delete logic from mixed_db_client.go

zbud-msft:
#### Why I did it

Race condition fixed in #375 in db_client.go. Same logic was copied which introduced issue in #139 was copied in #271. Removing it.

#### How I did it

Removed logic of adding delete filed to map and deleting it introduced in #139, coped in #271.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


qiluo-msft:
This is a disruptive feature reverting. Let's explore a proper bugfix.

ganglyu (approved):
LGTM

PR #434, opened 2025-07-02
Support CloseWrite event such that we don't reload on every Write event

zbud-msft:
#### Why I did it

MS ADO: 33377547

If we have a slow writer to a cert file that is monitored such that they truncate the file, and write bytes at a time, we will see excessive amount of WRITE events such that our server will try to reload on each one. With CloseWrite supported, once a writer has finished writing to the file and closed the file we will then try the reload. We also add logic to distinguish between CREATE and IN_MOVED_TO events such that we will reload on IN_MOVED_TO but not CREATE events.

#### How I did it

1) Added a patch for fsnotify to support CLOSE_WRITE events and distinguish between CREATE and IN_MOVED_TO events
2) Change telemetry server to monitor for CLOSE_WRITE instead of just WRITE events and monitor IN_MOVED_TO instead of CREATE events.

Patched is based off https://github.com/fsnotify/fsnotify/pull/252

#### How to verify it

Manual test, UT

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/telemetry/telemetry.go
+++ b/telemetry/telemetry.go
@@ -283,7 +283,7 @@ func iNotifyCertMonitoring(watcher *fsnotify.Watcher, telemetryCfg *TelemetryCon
 					filepath.Ext(event.Name) == ".cer" || filepath.Ext(event.Name) == ".pem" ||
 					filepath.Ext(event.Name) == ".key") {
 					log.V(1).Infof("Inotify watcher has received event: %v", event)
-					if event.Op&fsnotify.Write == fsnotify.Write || event.Op&fsnotify.Create == fsnotify.Create {
+					if event.Op&fsnotify.CloseWrite == fsnotify.CloseWrite || event.Op&fsnotify.Create == fsnotify.Create {
 						log.V(1).Infof("Cert File has been modified: %s", event.Name)
 						serverControlSignal <- ServerStart // let server know that a write/create event occurred
 						done <- true
--- a/telemetry/telemetry.go
+++ b/telemetry/telemetry.go
@@ -283,7 +283,7 @@ func iNotifyCertMonitoring(watcher *fsnotify.Watcher, telemetryCfg *TelemetryCon
 					filepath.Ext(event.Name) == ".cer" || filepath.Ext(event.Name) == ".pem" ||
 					filepath.Ext(event.Name) == ".key") {
 					log.V(1).Infof("Inotify watcher has received event: %v", event)
-					if event.Op&fsnotify.CloseWrite == fsnotify.CloseWrite || event.Op&fsnotify.Create == fsnotify.Create {
+					if event.Op&fsnotify.CloseWrite == fsnotify.CloseWrite || event.Op&fsnotify.MovedTo == fsnotify.MovedTo {
 						log.V(1).Infof("Cert File has been modified: %s", event.Name)
 						serverControlSignal <- ServerStart // let server know that a write/create event occurred
 						done <- true
--- a/telemetry/telemetry_test.go
+++ b/telemetry/telemetry_test.go
@@ -353,7 +353,10 @@ func copyFile(srcPath string, destPath string) error {
 
 func writeSlowKey(backupKeyPath string, keyPath string) error {
 	// Copy existing key from keyPath to backupKeyPath
-	copyFile(keyPath, backupKeyPath)
+	err := copyFile(keyPath, backupKeyPath)
+	if err != nil {
+		return err
+	}
 
 	// Write from backupKeyPath to keyPath
 	backupKey, err := os.Open(backupKeyPath)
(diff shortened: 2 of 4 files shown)


telemetry/telemetry.go
```diff
@@ -283,7 +283,7 @@ func iNotifyCertMonitoring(watcher *fsnotify.Watcher, telemetryCfg *TelemetryCon
 					filepath.Ext(event.Name) == ".cer" || filepath.Ext(event.Name) == ".pem" ||
 					filepath.Ext(event.Name) == ".key") {
 					log.V(1).Infof("Inotify watcher has received event: %v", event)
-					if event.Op&fsnotify.Write == fsnotify.Write || event.Op&fsnotify.Create == fsnotify.Create {
+					if event.Op&fsnotify.CloseWrite == fsnotify.CloseWrite || event.Op&fsnotify.Create == fsnotify.Create {
```
make1980, line 286: >event.Op&fsnotify.Create == fsnotify.Create [](http://example.com/codeflow?start=63&length=43)

if we trigger a reload when create signal is received wouldn't we have a race condition where the file is still being rewritten and we start reloading? it seems that when CREATE is received there will be a CLOSE_WRITE event later on as well so we may just wait for CLOSE_WRITE?

also if the file is moved from another location to the location that we're monitoring we should check IN_MOVED_TO event as well, right?

zbud-msft: 1) There is a potential timing issue where if the cert file is created and then written to that we could potentially miss the CLOSE_WRITE event thus reloading with the non-written or partial-written cert/key. I think the one way we can fix this is with a debounce timer such that we check for incoming events in the next couple of seconds and we use the latest event as the event to act on. We can't just use CLOSE_WRITE as the only option as there are certain operations where only CREATE is triggered but not CLOSE_WRITE such as when you execute "mv" from another location to the monitored directory.

2) CREATE/CLOSE_WRITE are triggered when we move /copy/write to the cert files in the monitored directory.

make1980: for mv it's actually rename() syscall invoked which triggers IN_MOVED_TO event:

kema2@KEMA2-SL1:/mnt/c/Users/kema2$ strace mv temp1.csv.bak temp1.csv
renameat(AT_FDCWD, "temp1.csv.bak", AT_FDCWD, "temp1.csv") = 0

[https://man7.org/linux/man-pages/man7/inotify.7.html](https://man7.org/linux/man-pages/man7/inotify.7.html)

Suppose an application is watching the directories dir1 and dir2,
       and the file dir1/myfile.  The following examples show some events
       that may be generated.

           link("dir1/myfile", "dir2/new");
                  Generates an IN_ATTRIB event for myfile and an
                  IN_CREATE event for dir2.

           rename("dir1/myfile", "dir2/myfile");
                  Generates an IN_MOVED_FROM event for dir1, an
                  IN_MOVED_TO event for dir2, and an IN_MOVE_SELF event
                  for myfile.  The IN_MOVED_FROM and IN_MOVED_TO events
                  will have the same cookie value.

make1980: btw I don't think we'll ever miss some notification as long as we consume the event as fast as possible.

zbud-msft: You are right, fsnotify however uses "CREATE" for both CREATE and IN_MOVED_TO calls: https://github.com/fsnotify/fsnotify/blob/c2828203cd70a50dcccfb2761f8b1f8ceef9a8e9/inotify.go#L321

	if mask&unix.IN_CREATE == unix.IN_CREATE || mask&unix.IN_MOVED_TO == unix.IN_MOVED_TO {
		e.Op |= Create
	}
	
Which explains why when I move a file to the monitored directory we still only just see the create event,

telemetry/telemetry_test.go
```diff
@@ -330,6 +331,64 @@ func saveCertKeyPair(certPath, keyPath string) error {
...
+
+func writeSlowKey(backupKeyPath string, keyPath string) error {
+	// Copy existing key from keyPath to backupKeyPath
+	err := copyFile(keyPath, backupKeyPath)
+	if err != nil {
+		return err
+	}
+
+	// Write from backupKeyPath to keyPath
+	backupKey, err := os.Open(backupKeyPath)
```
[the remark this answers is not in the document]
zbud-msft: Done

telemetry/telemetry_test.go
```diff
@@ -804,6 +863,68 @@ func TestINotifyCertMonitoringDeletion(t *testing.T) {
 	}
 }
 
+func TestINotifyCertMonitoringSlowWrites(t *testing.T) {
+	testServerCert := "../testdata/certs/testserver.cert"
+	testServerKey := "../testdata/certs/testserver.key"
+	testServerKeyBackup := "../testdata/testserver.key"
+	timeoutInterval := 30
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(timeoutInterval)*time.Second)
```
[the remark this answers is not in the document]
zbud-msft: Done

make1980 (approved):
:shipit:

PR #460, opened 2025-07-24
Ensure test signal is created after watcher is added

zbud-msft:
#### Why I did it

Recent flakiness with telemetry_test.go blocking pipelines

This flakiness is caused by file notify events happening before watcher is added to observe the events, therefore missing crucial events

With this change, we ensure that the testReadySignal is started after the watcher is added.

#### How I did it

Code change, UT

#### How to verify it

UT

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


hdwhdw (approved):
LGTM

PR #474, opened 2025-08-07
Simplify option support for all SHOW paths

zbud-msft:
#### Why I did it

With this PR, we achieve the following:

1) Each SHOW command can register some common options it wants to support 

```
	SHOW_CMD_OPT_PERIOD = sdc.NewShowCmdOption(
		"period",
		sdc.Optional,
		SHOW_CMD_OPT_PERIOD_DESC,
		sdc.IntValue,
	)

	SHOW_CMD_OPT_JSON = sdc.NewShowCmdOption(
		"json",
		sdc.Optional,
		SHOW_CMD_OPT_JSON_DESC,
		sdc.BoolValue,
	)

	SHOW_CMD_OPT_INTERFACE = sdc.NewShowCmdOption(
		"interface",
		sdc.Required,
		SHOW_CMD_OPT_INTERFACE_DESC,
		sdc.StringValue,
	)

	sdc.RegisterCliPath(
		[]string{"SHOW", "interface", "counters"},
		getInterfaceCounters,
		nil,
		SHOW_CMD_OPT_NAMESPACE,
		SHOW_CMD_OPT_DISPLAY,
		SHOW_CMD_OPT_INTERFACES,
		SHOW_CMD_OPT_PERIOD,
		SHOW_CMD_OPT_JSON,
		SHOW_CMD_OPT_VERBOSE,
	)
```

2) All validation logic for each path according to how the path was registered will be done in one central place. The validation will do the following
- Ensure all required options are present
- Return Unimplemented on options when an unimplemented option is provided
- Ensure unsupported options are errored out with InvalidArgument
- If help is set to True, a response that shows all registered options and subcommand description will be sent out
- Each getter will no longer take in a path/prefix for them to parse but instead a map which contains all passed values as a key value pair, already validated, and ready to consume. We can grab the value by accessing the map and the data type it was registered with.

```
func getInterfaceCounters(options sdc.OptionMap) ([]byte, error)

	if interfaces, ok := options["interfaces"].Strings(); ok {
		ifaces = interfaces
	}

	if periodValue, ok := options["period"].Int(); ok {
		takeDiffSnapshot = true
		period = periodValue
	}

```

#### How I did it

Implemented validation when the path is passed in depending on the registered path config

#### How to verify it

UT

```
python /root/gnxi/gnmi_cli_py/py_gnmicli.py -g -t  xxxxx-p 50051 -m get -x interface/counters[help=true] -xt SHOW

Performing GetRequest, encoding=JSON_IETF to xxxxxx  with the following gNMI Path
 ------------------------- 
 [elem {
  name: "interface"
}
elem {
  name: "counters"
  key {
    key: "help"
    value: "true"
  }
}
]
The GetResponse is below
-------------------------

{
  "options": {
    "display": "[display=all] Show internal interfaces [default: all]",
    "help": "[help=true]Show this message",
    "interfaces": "[interfaces=TEXT] Filter by interfaces name",
    "json": "[json=true] No-op since response is in json format",
    "namespace": "UNIMPLEMENTED",
    "period": "[period=INTEGER] Display statistics over a specified period (in seconds)",
    "verbose": "[verbose=true] Enable verbose output"
  },
  "subcommands": null
}
-------------------------
````

```
python /root/gnxi/gnmi_cli_py/py_gnmicli.py -g -t xxxxxx -p 50051 -m get -x interface/counters[namespace=all] -xt SHOW
Performing GetRequest, encoding=JSON_IETF to xxxxxxx with the following gNMI Path
 ------------------------- 
 [elem {
  name: "interface"
}
elem {
  name: "counters"
  key {
    key: "namespace"
    value: "all"
  }
}
]
GRPC error
 rpc error: code = Unimplemented desc = option namespace is unimplemented
```

```
python /root/gnxi/gnmi_cli_py/py_gnmicli.py -g -t xxxxx-p 50051 -m get -x interface/counters[foo=bar] -xt SHOW
Performing GetRequest, encoding=JSON_IETF to xxxxx  with the following gNMI Path
 ------------------------- 
 [elem {
  name: "interface"
}
elem {
  name: "counters"
  key {
    key: "foo"
    value: "bar"
  }
}
]
GRPC error
 rpc error: code = InvalidArgument desc = option foo for path elem:{name:"interface"} elem:{name:"counters" key:{key:"foo" value:"bar"}} is not a valid option
```

```
python /root/gnxi/gnmi_cli_py/py_gnmicli.py -g -t xxxx -p 50051 -m get -x interface/counters[period=foo] -xt SHOW 
Performing GetRequest, encoding=JSON_IETF to xxxxxxx  with the following gNMI Path
 ------------------------- 
 [elem {
  name: "interface"
}
elem {
  name: "counters"
  key {
    key: "period"
    value: "foo"
  }
}
]
GRPC error
 rpc error: code = InvalidArgument desc = option period expects an int (got foo), err: strconv.Atoi: parsing "foo": invalid syntax
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### CODE REVIEW

File: show_client/show_opts.go [hunk 1]
```diff
@@ -0,0 +1,66 @@
...
+		sdc.BoolValue,
+	)
+
+	SHOW_CMD_OPT_NAMESPACE = sdc.NewShowCmdOption(
+		"namespace",
+		sdc.Unimplemented,
+		SHOW_CMD_OPT_UNIMPLEMENTED_DESC,
+		sdc.StringValue,
+	)
+
```
yutongzhang-microsoft, line 31: For some transceiver command, we also have an option `-hw, --fetch-from-hardware`, which is also Unimplemented. Would you mind adding this option, or we can add it later in another PR.

show_client/show_opts.go
```diff
@@ -0,0 +1,80 @@
...
+	SHOW_CMD_OPT_DISPLAY = sdc.NewShowCmdOption(
```
BYGX-wcr, line 32: this should also be no-op since we are not implementing multi-asic support?

zbud-msft: Yes it is still a no-op, will add comment

PR #506, opened 2025-09-04
Create make-ci PR pipeline for sonic-gnmi

ryanzhu706:
#### Why I did it
To validate formatting, tidies dependencies, builds all components, performs static analysis (vet, lint), executes tests with coverage, and runs security checks via Semgrep. This helps catch issues early, enforce coding standards, and maintain a stable codebase across contributions.
#### How to verify it
Drafted pipeline sample - https://dev.azure.com/mssonic/build/_build?definitionId=3302&_a=summary

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/azure-pipelines/azure-pipelines.yml
+++ b/azure-pipelines/azure-pipelines.yml
@@ -1,0 +1,188 @@
+# Starter pipeline
+# Start with a minimal pipeline that you can customize to build and deploy your code.
+# Add steps that build, run tests, deploy, and more:
+# https://aka.ms/yaml
+
+trigger:
+  branches:
+    include:
+    - master
+    - 202???
+    - 201???
+
+pr:
+  branches:
+    include:
+    - master
+    - 202???
+    - 201???
+
+variables:
+  - name: BUILD_BRANCH
+    ${{ if eq(variables['Build.Reason'], 'PullRequest') }}:
+      value: $(System.PullRequest.TargetBranch)
+    ${{ else }}:
+      value: $(Build.SourceBranchName)
+  - name: UNIT_TEST_FLAG
+    value: 'ENABLE_TRANSLIB_WRITE=y'
+
+resources:
+  repositories:
+  - repository: sonic-mgmt-common
+    type: github
+    name: sonic-net/sonic-mgmt-common
+    endpoint: sonic-net
+    ref: refs/heads/$(BUILD_BRANCH)
+  - repository: sonic-swss-common
+    type: github
+    name: sonic-net/sonic-swss-common
+    endpoint: sonic-net
+    ref: refs/heads/$(BUILD_BRANCH)
+
+stages:
+- stage: Build
+  jobs:
+  - job:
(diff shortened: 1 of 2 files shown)


azure-pipelines/azure-pipelines.yml
hdwhdw, line 1: So is this moving the original pipeline for sonic-gnmi to here?

azure-pipelines/gnmi-make-ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - feature/*
+
+pr:
+  branches:
+    include:
+      - master
+      - feature/*
+
+pool:
+  vmImage: 'ubuntu-latest'
```
hdwhdw, line 15: Let's pin the version to 2022.04.

azure-pipelines/gnmi-make-ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+      build-essential \
+      golang-go \
+      libyang-dev \
+      libnl-3-dev \
+      libnl-genl-3-dev \
+      curl \
+      git
+    if ! command -v redis-server &> /dev/null; then
+      echo "redis-server not found, installing from source..."
+      sudo apt-get install -y redis
```
hdwhdw, line 39: Probably doesn't matter but this is not installing from source, right?

azure-pipelines/gnmi-make-ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+- script: |
+    sudo apt-get update
+    sudo apt-get install -y \
+      build-essential \
+      golang-go \
+      libyang-dev \
+      libnl-3-dev \
+      libnl-genl-3-dev \
+      curl \
+      git
```
hdwhdw, line 36: I am not sure we need all of these? People might be using this as the source of truth for setting up dev environment so we should try to minimize unnecessary dependencies.

hdwhdw: Also I don't think the current CI pipeline needs redis-server.. (although we might need it at some point).

azure-pipelines/gnmi-make-ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+    cd sonic-gnmi-standalone
+    make validate-format
+    make tidy
+    make build-all
+    make vet
+    make lint
+    make test
+    make test-coverage
+    make verify
+    make semgrep
```
hdwhdw, line 65: Actually I think we should run these using `make ci`. This serves as a reference for developers so they can do the same thing (instead of running all these command individually) on their local machine.
```
  - script: |
      cd sonic-gnmi-standalone
      echo "Running Continuous Integration checks..."
      ## echo "Hint: you can also run these locally..."
      make ci
    displayName: 'Run CI pipeline'
```

PR #553, opened 2025-12-13
[DRAFT] Fix server restarting after cert rotation

zbud-msft:

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/telemetry/telemetry.go
+++ b/telemetry/telemetry.go
@@ -38,6 +38,78 @@ const (
 	ServerRestart ServerControlValue = iota // 2
 )
 
+
+type CertCache struct {
+	cert       *tls.Certificate
+	caPool     *x509.CertPool
+	caPath     string
+	mu         sync.RWMutex
+	hasWatcher atomic.Bool
+}
+
+func (c *CertCache) Get() *tls.Certificate {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.cert
+}
+
+func (c *CertCache) Set(cert *tls.Certificate) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.cert = cert
+}
+
+func (c *CertCache) GetCAPool() *x509.CertPool {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.caPool
+}
+
+func (c *CertCache) SetCAPool(pool *x509.CertPool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.caPool = pool
+}
+
+func loadAndValidateCert(certPath, keyPath string) (*tls.Certificate, error) {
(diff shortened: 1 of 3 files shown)


hdwhdw (commented):
AI Agent on behalf of Dawei:

Thanks for working on this — the direction of hot-reloading certs via `CertCache` + `GetCertificate` instead of restarting the server is the right call. A few things below: one bug, one architectural suggestion, and some smaller items.

---

**Bug: `VerifyPeerCertificate` will never run when CA cert is configured**

With `ClientAuth: tls.RequireAndVerifyClientCert` and `ClientCAs` never set, Go's built-in TLS verification runs first against the **system root CAs**. If the client cert is signed by a custom CA (the whole point of `--ca_crt`), built-in verification fails and the handshake aborts *before* `VerifyPeerCertificate` is ever called.

From Go docs:
> If normal verification fails then the handshake will abort before considering this callback.

Fix: use `tls.RequireAnyClientCert` (requires cert, skips built-in verification) when you have a custom `VerifyPeerCertificate`. The `AllowNoClientCert` path (`RequestClientCert`) accidentally works fine since it's below `VerifyClientCertIfGiven`.

---

**Suggestion: extract the new cert code into `internal/certmgr`**

You're introducing `CertCache`, `loadAndValidateCert`, `loadAndValidateCA`, and reworking `iNotifyCertMonitoring` — that's a self-contained cert management domain. Since this is new code you're writing anyway, it would cost very little extra to put it in its own package rather than adding to the already-overloaded `telemetry` main package (~548 lines).

The repo already has an established `internal/` pattern (`internal/diskspace`, `internal/download`, `internal/hash`, etc.), so this fits naturally:

```
internal/certmgr/
  certmgr.go    # CertCache, LoadAndValidateCert, LoadAndValidateCA
  watcher.go    # WatchCerts (the reworked iNotifyCertMonitoring)
```

Benefits:
- `certmgr` has zero dependency on gRPC, server config, or gnmi — clean boundary
- Independently testable: cert loading/rotation tests don't need a gRPC server
- `telemetry.go` just imports `internal/certmgr` and uses its exported types
- Sets a pattern for gradually untangling the rest of the monolith

`telemetry.go` would still own the `tls.Config` construction and `GetCertificate` callback — it just delegates cert loading/caching/watching to `certmgr`.

---

**Smaller items:**
- `ServerStart` and `ServerRestart` constants are now dead code — nothing sends them
- `log.Fatalf` inside `GetCertificate` kills the entire process from a TLS handshake callback. Returning an error would just fail that one connection, which seems safer.
- The PR needs a rebase — master has added fields (`UnixSocket`, `CaCertLnk`, `CertzMetaFile`, etc.) that are missing here.

hdwhdw:
Closing the PR for inactivity. Feel free to reopen and tag for review if you still want to pursue it.

PR #570, opened 2026-02-03
Add --unix_socket flag for Unix Domain Socket support

hdwhdw:
Add `--unix_socket` flag to allow the telemetry server to listen on a Unix Domain Socket instead of TCP. This enables local-only access without TLS for clients like device-ops-agent.

#### Why I did it
To provide a secure, local-only gNMI endpoint that doesn't require TLS certificates. UDS is cleaner than TCP loopback because:
- No port conflicts
- File permission-based access control
- Only accessible locally by design

#### How I did it
- Added `UnixSocket` field to `gnmi_server.Config` struct
- Added `--unix_socket` CLI flag in telemetry.go
- Modified server bind logic: if `port <= 0` and `unix_socket` is set, listen on UDS
- Updated port validation to allow `port <= 0` when `--unix_socket` is specified

#### How to verify it
```bash
telemetry --port -1 --unix_socket /var/run/gnmi/gnmi.sock --noTLS --allow_no_client_auth
ls -la /var/run/gnmi/gnmi.sock
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,7 @@ require (
 	github.com/google/gnxi v0.0.0-20181220173256-89f51f0ce1e2
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
 	github.com/kylelemons/godebug v1.1.0
+	github.com/maruel/natural v1.1.1
 	github.com/msteinert/pam v0.0.0-20201130170657-e61372126161
 	github.com/openconfig/gnmi v0.0.0-20200617225440-d2b4e6a45802
 	github.com/openconfig/gnoi v0.3.0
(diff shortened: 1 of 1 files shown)


gnmi_server/server.go
```diff
@@ -163,7 +163,9 @@ type AuthTypes map[string]bool
 type Config struct {
```
zbud-msft: Please add UT, binding server to listen on different addresses, ex: localhost v4, v6, otherwise this lgtm

hdwhdw: Resolve as stale. Re purpose this PR to bind to UDS instead of ports.

PR #571, opened 2026-02-05
Fix GET path to decode JSON Pointer escaping (RFC 6901)

hdwhdw:
## Summary
Fixes gNMI paths failing to find keys that contain slashes (e.g., CIDR notation like `10.224.0.3/32`).

## Problem
The GET path was not decoding JSON Pointer escaping (RFC 6901) in table keys:
- `~1` should decode to `/`
- `~0` should decode to `~`

This caused GET to return empty results for keys written via SET, because:
- **SET path**: Decodes `~1` → `/` via `JsonClient.FixPath`, writes `VNET_ROUTE_TUNNEL|Vnet_123|10.224.0.3/32`
- **GET path**: Used raw path element `10.224.0.3~132`, looked for non-existent key

## Example
```bash
# SET writes with decoded key (works)
gnmic set --update-path "CONFIG_DB/localhost/VNET_ROUTE_TUNNEL/Vnet_123|10.224.0.3~132" ...

# GET couldn't find it (returned empty {})
gnmic get --path "CONFIG_DB/localhost/VNET_ROUTE_TUNNEL/Vnet_123|10.224.0.3~132" ...
# Returns: {}

# After fix: GET returns the data correctly
```

## Fix
Added shared `DecodeJsonPointer()` helper function (RFC 6901 compliant) and applied it to all affected code paths:

| File | Function | Change |
|------|----------|--------|
| `json_pointer.go` | `DecodeJsonPointer()` | NEW - shared helper for RFC 6901 decoding |
| `mixed_db_client.go` | `getDbtablePath()` | Decode `mappedKey` for table key lookups |
| `db_client.go` | `populateDbtablePath()` | Same fix for legacy DbClient |
| `show_client_helpers.go` | `checkOptionsInPath()` | Decode option values (e.g., CIDR in show commands) |

### Decode Order (RFC 6901 Section 4)
The order matters: `~1` → `/` first, then `~0` → `~`. This avoids incorrectly turning `~01` into `/` instead of `~1`.

## Fixes
ADO 36672792

## Test Plan
- [x] Unit test for `DecodeJsonPointer()` edge cases
- [x] Integration test for `getDbtablePath()` with encoded keys
- [ ] Verify GET returns correct data for keys containing `/`
- [ ] Verify show commands with CIDR arguments work correctly

### DIFF
--- a/sonic_data_client/client_test.go
+++ b/sonic_data_client/client_test.go
@@ -925,6 +925,13 @@ func TestGetDbtablePathJsonPointerDecoding(t *testing.T) {
 				t.Skip("Skipping: CONFIG_DB not found in RedisDbMap")
 			}
 
+			// Verify the Redis client actually exists before proceeding
+			// initRedisDbMap may have cleared the map and failed to repopulate
+			lookupKey := mapkey + ":CONFIG_DB"
+			if _, ok := RedisDbMap[lookupKey]; !ok {
+				t.Skipf("Skipping: Redis client for %q not present in RedisDbMap", lookupKey)
+			}
+
 			client := MixedDbClient{
 				prefix:        prefix,
 				target:        "CONFIG_DB",


sonic_data_client/mixed_db_client.go
```diff
@@ -673,6 +670,9 @@ func (c *MixedDbClient) getDbtablePath(path *gnmipb.Path, value *gnmipb.TypedVal
 	var mappedKey string
 	if len(stringSlice) > 2 { // tmp, to remove mappedKey
 		mappedKey = stringSlice[2]
+		// Decode JSON Pointer escaping (RFC 6901): ~1 -> /, ~0 -> ~
```
zbud-msft: Is this change applicable to any other data cleints, like db_client.go?

hdwhdw: No. Added the fix.

PR #581, opened 2026-02-17
convert updates_only to suppress_redundant

ronan-nexthop:
#### Why I did it
This flag was incorrectly implemented.

See related part of gNMI spec at https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#35152-stream-subscriptions

> Optionally, the suppress_redundant field of the Subscription message may be set for a sampled subscription. In the case that it is set to true, the target SHOULD NOT generate a telemetry update message unless the value of the path being reported on has changed since the last update was generated. Updates MUST only be generated for those individual leaf nodes in the subscription that have changed. 

On the other hand, upadtes_only effectively just means skip the initial sync and then send samples as per usual.

Note this will need changes to sonic-mgmt also, I am holding off on doing that until the gNMIC PR I have raised is merged as this will make things easier. Raising the PR now just to show it is a triaged issue.

#### How I did it
Made the same code flow look at the suppress redundant flag and not updates_only as it behaved in the way it should when the suppress_redundant flag is sent.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/gnmi_server/server_test.go
+++ b/gnmi_server/server_test.go
@@ -1119,9 +1119,10 @@ func prepareDbTranslib(t *testing.T) {
 
 // subscriptionQuery represent the input to create an gnmi.Subscription instance.
 type subscriptionQuery struct {
-	Query          []string
-	SubMode        pb.SubscriptionMode
-	SampleInterval uint64
+	Query             []string
+	SubMode           pb.SubscriptionMode
+	SampleInterval    uint64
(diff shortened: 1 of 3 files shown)


hdwhdw:
Closing as stale. Please reopen when the work is ready to resume.

PR #603, opened 2026-03-05
sonic_data_client: fix PollStats goroutine leak causing data race in tests

sigabrtv1-ui:
## Problem

Three related issues cause `TestProcessTransferContent_{OpenFileError,WriteError,CloseError}` to crash or fail silently on master CI:

### 1. PollStats goroutine leaks into test runs (nil pointer panic / DATA RACE)

`PollStats()` is started in `init()` in `sonic_data_client/non_db_client.go` and never stopped. When tests in `gnmi_server` use gomonkey to patch `os.OpenFile` globally, the still-running `PollStats` goroutine calls `linuxproc.ReadStat("/proc/stat")` which goes through `os.OpenFile` — hits the mock — and receives a zero-value `*os.File{}` back. Calling `Stat()` on that zero-value file triggers a nil pointer dereference:

```
panic: runtime error: invalid memory address or nil pointer dereference
goroutine in PollStats() → linuxproc.ReadStat → os.ReadFile → os.(*File).Stat
```

Also observed as a DATA RACE in CI build `20260305.14`.

### 2. glog background goroutine hits `os.OpenFile` mock → `os.Exit`

The three `TestProcessTransferContent_*` tests patch `os.OpenFile` globally via gomonkey. glog runs a background goroutine that calls `os.OpenFile` to rotate log files. It hits the mock and receives either an error or a zero-value `*os.File{}`, then calls `os.Exit(2)`, killing the test binary silently:

```
log: exiting because of error: log: cannot create log: simulated OpenFile failure
FAIL github.com/sonic-net/sonic-gnmi/gnmi_server  (no output)
```

## Fix

**`sonic_data_client/non_db_client.go`**
- Add `pollStatsDone chan struct{}` + `sync.Once`
- Add exported `StopPollStats()` — safe to call multiple times
- `PollStats()` selects on done channel each iteration and exits when closed
- Use `select { case <-time.After(...) }` instead of `time.Sleep` for immediate shutdown response

**`gnmi_server/server_test.go`**
- Call `sdc.StopPollStats()` in existing `TestMain` before running tests

**`gnmi_server/gnoi_os_test.go`**
- In each of the three `TestProcessTransferContent_*` tests: flush glog and set `logtostderr=true` before applying the `os.OpenFile` patch, restore in `defer`. This prevents glog's background goroutine from calling `os.OpenFile` during the window where it is patched.

## Test Results

```
$ go test -race -gcflags=all=-l -mod=vendor -run TestProcessTransferContent ./gnmi_server/

=== RUN   TestProcessTransferContent_OpenFileError
--- PASS: TestProcessTransferContent_OpenFileError (0.00s)
=== RUN   TestProcessTransferContent_WriteError
--- PASS: TestProcessTransferContent_WriteError (0.00s)
=== RUN   TestProcessTransferContent_CloseError
--- PASS: TestProcessTransferContent_CloseError (0.00s)
ok      github.com/sonic-net/sonic-gnmi/gnmi_server     7.620s
```

Note: `-gcflags=all=-l` is required (see PR #604) for gomonkey patches to take effect.

### DIFF
--- a/gnmi_server/gnoi_os_test.go
+++ b/gnmi_server/gnoi_os_test.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"crypto/tls"
 	"errors"
+	"flag"
 	"fmt"
 	"github.com/agiledragon/gomonkey/v2"
+	glog "github.com/golang/glog"
 	ospb "github.com/openconfig/gnoi/os"
 	"github.com/stretchr/testify/assert"
 	"google.golang.org/grpc"
@@ -739,11 +741,20 @@ func newTestOSServer() *OSServer {
 }
 
 func TestProcessTransferContent_OpenFileError(t *testing.T) {
+	// Flush and redirect glog before patching os.OpenFile. Without this, glog's
+	// background flush goroutine can call os.OpenFile to rotate its log file and
+	// hit our mock, causing glog to call os.Exit and crash the test binary.
+	glog.Flush()
+	flag.Set("logtostderr", "true")
+
 	// Mock the os.OpenFile function to simulate a failure
 	patch := gomonkey.ApplyFunc(os.OpenFile, func(name string, flag int, perm os.FileMode) (*os.File, error) {
 		return nil, errors.New("simulated OpenFile failure")
 	})
-	defer patch.Reset()
+	defer func() {
+		patch.Reset()
+		flag.Set("logtostderr", "false")
+	}()
 
 	srv := newTestOSServer()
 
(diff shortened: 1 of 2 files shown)


sigabrtv1-ui:
@zbud-msft @FengPan-Frank — would appreciate a review when you get a chance. This fixes the flaky DATA RACE on master (build 20260305.14, TestProcessTransferContent_CloseError) caused by the PollStats goroutine racing with gomonkey-patched os.OpenFile calls in gnmi_server tests.

PR #549, opened 2025-12-05
Implements the frontend logic for gNSI Authz

niranjanivivek:
This commit enables support for policy-based authorization of gRPC requests in gNMI server and handles the Authz Rotate() RPC for secure policy updates.

HLD Reference : [gNSI HLD](https://github.com/rlucus/SONiC/blob/c4c043118d7cf07eab687dc4085f4af1f9346770/doc/mgmt/gnmi/gnsi.md)

Dependency Chain for Merge

Please follow this merge order,

Backend Base PR: sonic-mgmt-common #https://github.com/sonic-net/sonic-mgmt-common/pull/183 (Must be merged first)
Backend Incremental PR: sonic-mgmt-common #https://github.com/sonic-net/sonic-mgmt-common/pull/200
Current PR: sonic-gnmi #https://github.com/sonic-net/sonic-gnmi/pull/549

**gNSI Authz - UT Results**
=== RUN   TestGnsiAuthzRotation
=== RUN   TestGnsiAuthzRotation/RotateOpenClose
=== RUN   TestGnsiAuthzRotation/RotateStreamRecvError
=== RUN   TestGnsiAuthzRotation/RotateStreamSendError
=== RUN   TestGnsiAuthzRotation/RotatePolicyEmptyRequest
=== RUN   TestGnsiAuthzRotation/RotatePolicyEmptyUploadRequest
=== RUN   TestGnsiAuthzRotation/RotatePolicyWrongJSON
=== RUN   TestGnsiAuthzRotation/RotatePolicyNoVersion
=== RUN   TestGnsiAuthzRotation/RotatePolicySuccess
=== RUN   TestGnsiAuthzRotation/RotatePolicyNoFinalize
=== RUN   TestGnsiAuthzRotation/RotateTheSamePolicyTwice
=== RUN   TestGnsiAuthzRotation/RotateTheSamePolicyTwiceWithForceOverwrite
=== RUN   TestGnsiAuthzRotation/ParallelRotationCalls
--- PASS: **TestGnsiAuthzRotation** (4.38s)
    --- PASS: TestGnsiAuthzRotation/RotateOpenClose (0.02s)
    --- PASS: TestGnsiAuthzRotation/RotateStreamRecvError (0.02s)
    --- PASS: TestGnsiAuthzRotation/RotateStreamSendError (0.01s)
    --- PASS: TestGnsiAuthzRotation/RotatePolicyEmptyRequest (0.02s)
[...]

PR #604, opened 2026-03-05
Makefile: add -gcflags=all=-l to gnmi_server tests for gomonkey compatibility

sigabrtv1-ui:
## Problem

Tests in `gnmi_server` that use `gomonkey.ApplyFunc()` to mock OS/auth functions (e.g. `user.Lookup`, `UserPwAuth`) silently fail to patch because the binary is built with inlining enabled. gomonkey patches function pointers at the machine-code level and **requires** inlining to be disabled to work correctly.

Without `-gcflags=all=-l`, `ApplyFunc()` patches are no-ops at runtime — the real functions execute instead — causing tests to get real authentication failures:

```
E0305 14:42:56 gnsi_authz.go:66] authentication failed in Rotate RPC: rpc error: code = Unauthenticated desc = Unauthenticated
    gnsi_authz_test.go:60: Unexpected error: rpc error: code = Unauthenticated desc = Unauthenticated
--- FAIL: TestGnsiAuthzRotation/RotateOpenClose
--- FAIL: TestGnsiAuthzRotation/RotateStreamRecvError
... (all 11 subtests fail)
--- FAIL: TestGnsiAuthzRotation
```

This is blocking PR #549 (gNSI Authz frontend implementation) which has failed on every CI run.

## Fix

Add `-gcflags=all=-l` to the `go test` invocations for `gnmi_server` (in `check_gotest`) and `INTEGRATION_ENV_PKGS` (in `check_gotest_junit`). This flag disables inlining so gomonkey can insert its patches at the function entry points.

The flag has no effect on production binaries and does not break any other tests — it only makes the test binary slightly larger and marginally slower to compile.

## References

- [gomonkey docs: build flags requirement](https://github.com/agiledragon/gomonkey#notice)
- PR #549: gNSI Authz (blocked by this issue)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,7 +222,7 @@ $(ENVFILE):
 check_gotest: $(DBCONFG) $(ENVFILE)
 	sudo CGO_LDFLAGS="$(CGO_LDFLAGS)" CGO_CXXFLAGS="$(CGO_CXXFLAGS)" $(GO) test -race -coverprofile=coverage-telemetry.txt -covermode=atomic -mod=vendor -v github.com/sonic-net/sonic-gnmi/telemetry
 	sudo CGO_LDFLAGS="$(CGO_LDFLAGS)" CGO_CXXFLAGS="$(CGO_CXXFLAGS)" $(GO) test -race -coverprofile=coverage-config.txt -covermode=atomic -v github.com/sonic-net/sonic-gnmi/sonic_db_config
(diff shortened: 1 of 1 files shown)


sigabrtv1-ui:
@zbud-msft @FengPan-Frank — companion to #603. Adds -gcflags=all=-l to gnmi_server test invocations so gomonkey patches work correctly. Without this flag, ApplyFunc patches are silently no-ops, causing auth mocks to be ignored and every test to fail with Unauthenticated.

PR #607, opened 2026-03-06
azure-pipelines: add nightly scheduled run for flakiness detection

sigabrtv1-ui:
## Problem

Test analytics in Azure DevOps accumulate results mostly from PR runs. PR run failures are ambiguous — a test might fail because the PR broke it, or because it's flaky. Without regular master runs, the analytics can't cleanly identify flaky tests.

With the current volume of master runs (only on merges, a few per day at peak), a test that flakes at 5–10% failure rate can go undetected for days or weeks.

## Solution

Add a nightly scheduled trigger on `master` with `always: true`, so test analytics gets a consistent stream of clean data points independent of merge activity.

- **Scheduled**: midnight UTC daily
- **`always: true`**: fires even with no new commits — critical for consistent data
- **`RUN_TYPE` variable**: tags test run titles as `Nightly` vs `CI` so scheduled and PR/push results can be filtered separately in Azure DevOps test analytics

## Impact

30 clean master data points per month, enough to reliably surface tests failing >10% of the time. Scheduled and PR results remain in the same analytics dashboard but are distinguishable by run title prefix.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -17,6 +17,17 @@ pr:
     - 202???
     - 201???
 
+# Scheduled nightly run on master for flakiness detection.
+# always: true ensures the run fires even when there are no new commits,
+# giving a consistent stream of data points for test analytics.
+schedules:
+- cron: "0 0 * * *"
+  displayName: "Nightly master run"
+  branches:
+    include:
+    - master
+  always: true
+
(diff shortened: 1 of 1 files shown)


hdwhdw:
Closing as stale. Please reopen when the work is ready to resume.

PR #609, opened 2026-03-06
[RFC] pkg/bypass: use go-redis/v7 instead of go-redis/v9

sigabrtv1-ui:
## ⚠️ This is NOT for merge — demo/RFC only

Showing colleagues what's required to downgrade `pkg/bypass` from `github.com/redis/go-redis/v9` to `github.com/go-redis/redis/v7`.

---

## What changes

| Operation | v9 (current) | v7 |
|---|---|---|
| Import | `redis/go-redis/v9` | `go-redis/redis/v7` |
| HGet | `HGet(ctx, key, field)` | `HGet(key, field)` |
| HSet (map) | `HSet(ctx, key, map)` | `HMSet(key, map)` |
| HSet (field) | `HSet(ctx, key, f, v)` | `HSet(key, f, v)` |
| Del | `Del(ctx, key)` | `Del(key)` |

The main difference is v7 doesn't take `context.Context` in Redis commands, and uses `HMSet` for map writes.

`go-redis/v7` is already an indirect dep in go.mod, so no new dependency is introduced — just promoted to direct.

## Tests

All existing `pkg/bypass` tests pass with zero changes to test logic.

### DIFF
--- a/pkg/bypass/bypass.go
+++ b/pkg/bypass/bypass.go
@@ -9,9 +9,9 @@ import (
 	"os"
 	"strings"
 
+	"github.com/go-redis/redis/v7"
 	"github.com/golang/glog"
 	gnmipb "github.com/openconfig/gnmi/proto/gnmi"
-	"github.com/go-redis/redis/v7"
 	"google.golang.org/grpc/metadata"
 )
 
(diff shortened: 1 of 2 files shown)


sigabrtv1-ui:
Closing — the team has decided to move away from gomonkey going forward, so the Redis v7/v9 bypass workaround is no longer needed.

PR #620, opened 2026-03-17
Fix too_many_pings during DPU SetPackage by coordinating keepalive settings

sigabrtv1-ui:
## What I did
Fixed a keepalive mismatch between the DPU proxy client and gNMI server that causes `ENHANCE_YOUR_CALM` / `"too_many_pings"` disconnections during long-running `SetPackage` operations on smartswitch DPUs.

## How I did it
1. **Proxy client** (`pkg/interceptors/dpuproxy/proxy.go`): Increased keepalive ping interval from 10s → 30s and timeout from 3s → 10s.
2. **gNMI server** (`telemetry/telemetry.go`): Added explicit `KeepaliveEnforcementPolicy` with `MinTime: 20s` so the server tolerates the proxy ping frequency instead of using the default 5-minute minimum.

## Why I did it
During DPU image installation via `SetPackage`, `sonic-installer install` can run for several minutes. The proxy client was pinging every 10s, but the DPU server's default enforcement policy (MinTime=5m) treated this as a flood, terminating the connection with `too_many_pings`.

The 30s client interval with 20s server minimum provides a safe margin while still detecting dead connections reasonably fast.

## How to verify it
1. Deploy on a smartswitch with DPU
2. Run `SetPackage` gNOI RPC targeting a DPU with a valid image
3. Verify the installation completes without `too_many_pings` disconnection
4. Verify keepalive still detects dead connections (kill the DPU gNMI server mid-operation)

Fixes: #619
Related: sonic-net/sonic-mgmt#23054


ryanzhu706 (dismissed):

sigabrtv1-ui (commented):
Addressed review feedback: keepalive values are now configurable via environment variables (GNMI_CLIENT_KEEPALIVE_INTERVAL, GNMI_CLIENT_KEEPALIVE_TIMEOUT, GNMI_SERVER_KEEPALIVE_MIN_TIME). Defaults remain the same so behavior is unchanged out of the box.

vaibhavhd (dismissed):
Approving based on offline discussion about a follow up PR where the timeouts/settings will be introduced as config.

hdwhdw (dismissed):

PR #632, opened 2026-03-25
[agent][clientCertAuth] Fix nil pointer dereference on HTTP response in TryDownload

### LINKED ISSUE #630: [clientCertAuth] Nil pointer dereference on HTTP response when http.Get() fails
### Description

In `gnmi_server/clientCertAuth.go` around line 161, `resp.StatusCode` is accessed even when `http.Get()` returns `(nil, err)`. The current code checks `resp != nil` for the defer but then unconditionally accesses `resp.StatusCode` in the error check.

### Current code
```go
resp, err := http.Get(url)
if resp != nil { defer resp.Body.Close() }
if err != nil || resp.StatusCode != http.StatusOK {  // resp may be nil here!
```

### Fix
Check `err != nil` first, return early, then access `resp`:
```go
resp, err := http.Get(url)
if err != nil {
    return err
}
defer resp.Body.Close()
if resp.StatusCode != http.StatusOK {
```

### Impact
Panic in the gNMI server when client certificate authentication makes an HTTP call that fails at the network level.

rustiqly:
#### What I did
Fix panic in `TryDownload()` when `http.Get()` returns `(nil, err)`.

#### How I did it
Restructured error handling to check `err != nil` first and return early before accessing `resp`. Then `defer resp.Body.Close()` and check `resp.StatusCode` separately.

**Before:** `resp.StatusCode` accessed in combined check with `err`, panics when `resp` is nil.
**After:** Error checked first, `resp` only accessed after confirming non-nil.

#### How to verify it
Call `TryDownload()` with an unreachable URL — should return false instead of panic.

#### Which release branch to backport
master

Fix nil pointer panic in gNMI CRL download when HTTP request fails.

Fixes: #630

### DIFF
--- a/gnmi_server/clientCertAuth.go
+++ b/gnmi_server/clientCertAuth.go
@@ -161,6 +161,9 @@ func TryDownload(url string) bool {
 	resp, err := http.Get(url)
 
 	if err != nil {
+		if resp != nil && resp.Body != nil {
+			resp.Body.Close()
+		}
 		glog.Infof("Download CRL: %s failed: %v", url, err)
 		return false
 	}
--- a/gnmi_server/clientCertAuth.go
+++ b/gnmi_server/clientCertAuth.go
@@ -160,12 +160,14 @@ func TryDownload(url string) bool {
 	glog.Infof("Download CRL start: %s", url)
 	resp, err := http.Get(url)
 
-	if resp != nil {
-		defer resp.Body.Close()
-	}
-
-	if err != nil || resp.StatusCode != http.StatusOK {
+	if err != nil {
 		glog.Infof("Download CRL: %s failed: %v", url, err)
+		return false
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		glog.Infof("Download CRL: %s failed: HTTP %d", url, resp.StatusCode)
 		return false
 	}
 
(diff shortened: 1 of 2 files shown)


gnmi_server/clientCertAuth.go
```diff
@@ -160,12 +160,17 @@ func TryDownload(url string) bool {
...
-	if resp != nil {
-		defer resp.Body.Close()
+	if err != nil {
+		if resp != nil && resp.Body != nil {
+			resp.Body.Close()
+		}
+		glog.Infof("Download CRL: %s failed: %v", url, err)
+		return false
 	}
+	defer resp.Body.Close()
```
yxieca: 1. There is no nil check here for resp or resp.Body
2. If we move this line above line 163, can line163 body be a clean return?

yxieca:
@rustiqly please consider adding unit tests.

rustiqly:
@yxieca acknowledged, I will work on adding unit tests for this in the next iteration.

rustiqly:
Addressed the latest review feedback in `1c19b11`:\n\n- Moved CRL response body closing to a single deferred close guarded by `resp != nil && resp.Body != nil`.\n- Added tests for redirect-error and non-OK response paths.\n\nLocal test note: `go test ./gnmi_server -run 'TestTryDownload' -count=1` is blocked on this host because the installed Go toolchain is 1.22.2 while `github.com/openconfig/gnsi@v1.9.0` requires Go >= 1.23.0.

hdwhdw (approved):

hdwhdw (approved):

PR #634, opened 2026-03-31
Fix too_many_pings during DPU SetPackage and add ARM64 CI build

hdwhdw:
## What I did
Fixed a keepalive mismatch between the DPU proxy client and gNMI server that causes `ENHANCE_YOUR_CALM` / `too_many_pings` disconnections during long-running `SetPackage` operations on smartswitch DPUs. Also added ARM64 deb build support to the Azure CI pipeline.

### Keepalive fix
1. **Proxy client** (`pkg/interceptors/dpuproxy/proxy.go`): Increased keepalive ping interval from 10s → 30s and timeout from 3s → 10s.
2. **gNMI server** (`telemetry/telemetry.go`): Added explicit `KeepaliveEnforcementPolicy` with `MinTime: 20s` so the server tolerates the proxy ping frequency instead of using the default 5-minute minimum.

### ARM64 CI build
- New shared `.azure/templates/build-deb.yml` template (arch-parameterized)
- Unified `.azure/templates/install-dependencies.yml` for both amd64 and arm64
- New `BuildAmd64` and `BuildArm64` pipeline stages running in parallel
- Separated deb packaging from integration test job to avoid duplicate builds

## Why I did it
During DPU firmware upgrade via `SetPackage`, `sonic-installer install` can run for several minutes. The proxy client was pinging every 10s, but the DPU server's default enforcement policy (MinTime=5m) treated this as a flood, terminating the connection with `too_many_pings`.

The 30s client interval with 20s server minimum provides a safe margin while still detecting dead connections reasonably fast.

ARM64 build support is needed for DPU platforms that run on ARM processors.

## How to verify it
1. Deploy on a smartswitch with DPU
2. Run `SetPackage` gNOI RPC targeting a DPU with a valid image
3. Verify the installation completes without `too_many_pings` disconnection
4. Verify keepalive still detects dead connections (stop the DPU gNMI server mid-operation)
5. Confirm ARM64 CI stage passes in Azure DevOps pipeline

Fixes: #619
Related: sonic-net/sonic-mgmt#23054

### DIFF
--- a/pkg/interceptors/dpuproxy/proxy.go
+++ b/pkg/interceptors/dpuproxy/proxy.go
@@ -177,13 +177,18 @@ func (p *DPUProxy) getConnection(ctx context.Context, dpuIndex, ipAddress string
 		target := fmt.Sprintf("%s:%s", ipAddress, port)
 		glog.Infof("[DPUProxy] Trying to connect to DPU%s at %s (attempt %d/%d)", dpuIndex, target, i+1, len(portsToTry))
 
-		// Create connection with keepalive settings for long-lived connections
+		// Create connection with keepalive settings for long-lived connections.
+		// Use a conservative ping interval to avoid triggering the server's
+		// default EnforcementPolicy (MinTime=5m). Operations like SetPackage
+		// can block for minutes during image installation; aggressive pinging
+		// causes the server to send GOAWAY with "too_many_pings".
+		// See: https://github.com/sonic-net/sonic-gnmi/issues/619
 		conn, err := grpc.NewClient(
 			target,
 			grpc.WithTransportCredentials(insecure.NewCredentials()),
 			grpc.WithKeepaliveParams(keepalive.ClientParameters{
-				Time:                10 * time.Second, // Send keepalive ping every 10s
-				Timeout:             3 * time.Second,  // Wait 3s for ping ack before considering connection dead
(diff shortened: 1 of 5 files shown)


pkg/interceptors/dpuproxy/proxy.go
```diff
@@ -177,15 +177,25 @@ func (p *DPUProxy) getConnection(ctx context.Context, dpuIndex, ipAddress string
 		target := fmt.Sprintf("%s:%s", ipAddress, port)
 		glog.Infof("[DPUProxy] Trying to connect to DPU%s at %s (attempt %d/%d)", dpuIndex, target, i+1, len(portsToTry))
 
-		// Create connection with keepalive settings for long-lived connections
+		// Create connection with keepalive settings for long-lived connections.
```
vaibhavhd: A `TODO` can help to track the future work for a config knob to control window size and timeouts

ryanzhu706 (dismissed):

vaibhavhd (commented):
It would be nicer to split this PR into two, per the two unique issues that are solved here - 1) too many pings 2) too much memory consumption

vaibhavhd (approved):
Non blocking changes suggested from last iteration,

vmittal-msft:
@hdwhdw please raise manual pr for 202511 due to conflicts

PR #644, opened 2026-04-10
Support stream multiplexing

zbud-msft:
#### Why I did it

Our gRPC server should be able to support several RPCs on a single TCP connection using HTTP/2 multiplexing, which is how gRPC is designed to work.

Check for UseSingleTcp flag to enable this feature. If enabled, we can support multiple rpcs on a single TCP connection regardless of request type (GET, POLL, SUBSCRIBE), target (Any DB), etc.

#### How I did it

Added a unique atomic ID to each Client instance.

#### How to verify it

UT/CI/Manual test

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/gnmi_server/client_subscribe.go
+++ b/gnmi_server/client_subscribe.go
@@ -15,11 +15,20 @@ import (
 	"sync/atomic"
 )
 
+// ClientKey is the key used in the server's client map for duplicate detection.
+// When EnableStreamMultiplexing is true, each stream gets a unique StreamID
+// so multiple streams from the same peer coexist. When false, StreamID is 0
+// and the peer address alone identifies the client (legacy behavior).
+type ClientKey struct {
+	PeerAddr string
+	StreamID uint64
+}
+
 // Client contains information about a subscribe client that has connected to the server.
 type Client struct {
 	addr         net.Addr
 	id           uint64 // Unique ID for this client instance
-	useSingleTcp bool   // When true, include unique ID in String() to allow multiple streams per TCP connection
+	enableStreamMultiplexing bool   // When true, include unique stream ID in Id() for multiplexing
 	sendMsg      int64
 	recvMsg      int64
 	errors       int64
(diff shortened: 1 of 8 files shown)


gnmi_server/server.go
```diff
@@ -233,6 +233,7 @@ type Config struct {
 	PathzPolicy     bool   // Enable gNMI pathz policy.
 	PathzPolicyFile string // Path to gNMI pathz policy file.
 	PathzMetaFile   string // Path to JSON file with pathz metadata.
+	UseSingleTcp    bool   // Allow multiple Subscribe RPCs on a single TCP connection.
```
hdwhdw: "EnableStreamMultiplexing" describe what you want to do better. UseSingleTcp is a bit vague.

gnmi_server/client_subscribe.go
```diff
@@ -65,7 +76,11 @@ func (c *Client) setConnectionManager(threshold int) {
 }
 
 // String returns the target the client is querying.
+// When useSingleTcp is enabled, includes a unique ID to allow multiple streams per TCP connection.
 func (c *Client) String() string {
```
hdwhdw: Using String() as a defacto id can cause a lot of problem in the future:
1. String() is mostly used for logging, and people change it all the time. Today we can have a client 10.0.0.1:12345 but if someone changes to client[10.0.0.1:12345] it breaks all your logic.

2. Most people will think String() give you a description who the client is, so its value should not depend on a server side flag. A client's identity become config dependent can lead to major confusion.

3. Parsing can be brittle.

4. Comparing clients (like whether they are equal, whether they are from same address, etc) can easily produce bugs. 

5. You can easily leak client if some state of the client changed after you create and before you remove.

Strongly suggest using proper struct for client identifier, such as 
```go
type ClientKey struct {
    PeerAddr  string
    StreamID  uint64
}
```

hdwhdw (changes requested):

gnmi_server/server.go
zbud-msft: Done.

gnmi_server/client_subscribe.go
zbud-msft: Done. This is legacy behavior using c.String() as an ID for connections inside the clients map that this PR will now address.

hdwhdw (dismissed):

hdwhdw (approved):

PR #651, opened 2026-04-17
Implement ConfigReload on full config SET update

### LINKED ISSUE #22362: Enhancement: Implement ConfigReload for full config update
### Is it platform specific

generic

### Importance or Severity

Medium

### Description of the enhancement

gNMI accepts SET requests with a full config delete and full config update, and treats that as a full config reload. This used to be implemented internally to gNMI. The config would be written to a temp file locally in the gNMI container, and when a Reboot was triggered a `config reload` would be called on that file. Recent changes have been made to implement Reboot at the system level, and this approach no longer works. The sonic-mgmt test case has been disabled, and this functionality has effectively been dropped.

The handler should be re-written to trigger ConfigReload directly through D-Bus, and decouple it from Reboot. This will be a change from the previous behavior, where the command only writes the config and a later Reboot triggers it; now the ConfigReload will be triggered immediately and the command will not complete until it returns. This is in line with how ConfigReload works on the device, and should not be much of a change. Additionally, we expect minimal disruption as the command is currently non-functional, so any users will have breaking upstream changes as it is. 

### Current Behavior

Currently, this command is not functional.

### Proposed Behavior

Sending a gNMI SET-UPDATE with a full config delete path and a full config update path will trigger a call to ConfigReload(), causing configuration to be reloaded immediately.

donggyu-nexthop:
#### Why I did it

Resolves https://github.com/sonic-net/sonic-buildimage/issues/22362

The current "full config update" operation over gNMI silently fails. 

This is because the operation relied on GNOI System Reboot to pick up the requested configuration and apply it; however, the proper reboot call implementation ripped out the configuration file handling to better align with the GNOI system specs. This left the `config reload` operation non-functional.

#### How I did it

`client.MixedDbClient` has a new field which can optionally store "post-response callbacks", which are just functions that will be invoked after the gRPC response has been issued back to the client.

During `SetFullConfig`, this field will be set to a callback that invokes `DBusClient::ConfigReload()`. Then, at the gNMI server layer, the server will invoke `MixedDbClient::RunCallback()` to execute the callback (if registered).

The reason this step had to be deferred until the very end of the request lifecycle was because the `ConfigReload` D-Bus endpoint will restart ALL services, including the gNMI server, which means that calling this function directly from within `SetFullConfig` leaves the gNMI client hanging with no response.

#### How to verify it
- prep a testbed with changes from this branch
- ssh into testbed, and use `gnmic` to issue SET commands over gnmi
    - example query:
```sh
gnmic -a "<DUT ip>:<DUT port>" --insecure set \
--encoding json_ietf \
--delete "sonic-db:/CONFIG_DB/localhost/" \
--update-path "sonic-db:/CONFIG_DB/localhost/" \
--update-file <path to json with new config db>
```

- then on the DUT, verify the changes were successful

```sh
# if you modified the description of ethernet 4 port
sonic-db-cli hget "PORT|Ethernet4" description
```

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

Implement gNMI config reload without relying on gNOI reboots

#### Link to config_db schema for YANG module changes

N/A for this PR

### DIFF
--- a/common_utils/notification_producer.go
+++ b/common_utils/notification_producer.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"time"
 
 	sdcfg "github.com/sonic-net/sonic-gnmi/sonic_db_config"
 
@@ -50,7 +51,10 @@ func IsAutoRestartEnabled() bool {
 		log.Errorf("failed to connect to Redis for auto_restart check: %v", err)
 		return false
 	}
-	value, err := redisDb.HGet("FEATURE|gnmi", "auto_restart").Result()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+	value, err := redisDb.HGet(ctx, "FEATURE|gnmi", "auto_restart").Result()
 	if err != nil {
 		log.Errorf("failed to read FEATURE|gnmi auto_restart: %v", err)
 		return false
--- a/common_utils/notification_producer.go
+++ b/common_utils/notification_producer.go
@@ -44,6 +44,20 @@ func GetRedisDBClient() (*redis.Client, error) {
 	return rclient, nil
 }
 
+func IsAutoRestartEnabled() bool {
+	redisDb, err := GetRedisDBClient()
+	if err != nil {
+		log.Errorf("failed to connect to Redis for auto_restart check: %v", err)
+		return false
+	}
+	value, err := redisDb.HGet("FEATURE|gnmi", "auto_restart").Result()
+	if err != nil {
+		log.Errorf("failed to read FEATURE|gnmi auto_restart: %v", err)
+		return false
+	}
+	return value == "enabled"
+}
+
 // NotificationProducer provides utilities for sending messages using notification channel.
 // NewNotificationProducer must be called for a new producer.
 // Close must be called when finished.
--- a/sonic_data_client/mixed_db_client.go
+++ b/sonic_data_client/mixed_db_client.go
@@ -96,6 +96,8 @@ type MixedDbClient struct {
 	synced sync.WaitGroup  // Control when to send gNMI sync_response
 	w      *sync.WaitGroup // wait for all sub go routines to finish
 	mu     sync.RWMutex    // Mutex for data protection among routines for DbClient
+
+	configReloadCallback func(autoRestartEnabled bool) // callback to issue after sending RPC response back to the gNMI client
 }
 
 // redis client connected to each DB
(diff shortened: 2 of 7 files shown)


ronan-nexthop:
@donggyu-nexthop can you use `git blame` and tag people who may be suitable to review here. I think because no reviewers are auto-assigned people may not be activly looking at this.

@sneelam20 perhaps a good example of why the @gnmi-dev alias would be good to have auto-added to this, or alternatively, some form of auto-assigning based on `git blame`

donggyu-nexthop:
@ganglyu @abdosi @jipanyang @ndas7 could you give this a look? thank you!

anders-nexthop:
@sneelam20 @hdwhdw can we get some reviewers assigned to look at this? I do not have permissions to add reviewers.

This PR was discussed in the 5/14 UMF meeting. sonic-gnmi used to support a whole-file config replace option, which was invoked by passing the filename to the rpc as an UPDATE. The original implementation relied on an early version of GnoiReboot which did a 'config reload' rather than a hard reboot. This feature broke when an actual GnoiReboot impl was added. This PR plumbs the feature correctly, using DBUS to call 'config reload' directly.

This is not the same usage as the [save-on-set](https://github.com/rlucus/SONiC/blob/master/doc/mgmt/gnmi/save_on_set.md) feature, which triggers 'config save' after every write to make config changes persistent. This approach is intended for whole-config operations instead, where the user wants to bulk-load a config throw away anything pre-existing, then reload the system so that configuration takes effect.

common_utils/notification_producer.go
```diff
@@ -44,6 +45,23 @@ func GetRedisDBClient() (*redis.Client, error) {
...
+func IsAutoRestartEnabled() bool {
+	redisDb, err := GetRedisDBClient()
+	if err != nil {
+		log.Errorf("failed to connect to Redis for auto_restart check: %v", err)
+		return false
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+	value, err := redisDb.HGet(ctx, "FEATURE|gnmi", "auto_restart").Result()
```
hdwhdw, line 57: `GetRedisDBClient()` routes this lookup to `STATE_DB`, but `FEATURE|gnmi:auto_restart` is in `CONFIG_DB`.

sonic_data_client/mixed_db_client.go
```diff
@@ -1452,13 +1454,31 @@ func (c *MixedDbClient) SetFullConfig(delete []*gnmipb.Path, replace []*gnmipb.U
...
 	}
+	defer os.Remove(fileName)
 
 	PyCodeInGo := fmt.Sprintf(PyCodeForYang, fileName)
 	err = RunPyCode(PyCodeInGo)
 	if err != nil {
 		return fmt.Errorf("Yang validation failed!")
 	}
 
+	c.configReloadCallback = func(autoRestartEnabled bool) {
```
hdwhdw, line 1465: `Set` returns success before `ConfigReload` runs. If reload fails after CONFIG_DB is flushed, the failure cannot be reported and there is no rollback. Is that an acceptable risk for the first version? If not, a higher-level option is to stage a checkpoint and let the SONiC host service apply or restore it after gNMI shuts down.

anders-nexthop: yeah this is generally the trick about this feature. `config reload` will bring down gnmi - how do we ensure we can send a clean response back to the user if the signal we are waiting for is gnmi container being restarted? we thought we could at least send a response that the command was accepted (passed yang validation and is being triggered). @hdwhdw do you have any ideas of a better way to handle this?

hdwhdw: This probably has to be delegated to host service, i.e. make that api transactional, so it either succeed or do nothing. For gnmi we might have to accept the risk of not knowing whether the operaiton is successful.

PR #652, opened 2026-04-19
telemetry: add --bind_address flag to restrict TCP listener

xq9mend:
Add `--bind_address` flag to the telemetry binary. When set, the gRPC TCP listener binds to the specified address instead of all interfaces.

This enables restricting the cleartext gRPC endpoint (`--noTLS`) to localhost only by passing `--bind_address 127.0.0.1`, preventing remote access to the unencrypted endpoint on devices without certs.

Default behavior (empty `--bind_address`) is unchanged: binds all interfaces.

VRF-aware listener is not affected (uses createVrfListener which handles its own binding).


xq9mend:
@zbud-msft @hdwhdw

hdwhdw (approved):

PR #655, opened 2026-04-21
[202511] Fix too_many_pings during DPU SetPackage and add ARM64 CI build (#634)

hdwhdw:
Cherry-pick of #634 to 202511.

The target PR touches files that were restructured on master after 202511 branched, so several prerequisite commits are bundled to avoid messy manual conflict resolution. Each prerequisite is a squashed merge commit from master.

## Commits

| # | PR | Purpose |
|---|----|---------|
| 1 | #557 | CI pipeline refactor — introduces `.azure/templates/install-dependencies.yml` |
| 2 | #568 | gNMI Set bypass validation fast path (touches azure-pipelines.yml) |
| 3 | #579 | UDS listener — renames `opts` → `commonOpts`/`tlsOpts` in telemetry.go |
| 4 | #597 | Inline diff coverage check in Azure pipeline |
| 5 | #634 | **Target fix** — too_many_pings keepalive fix + ARM64 CI build |

With prerequisites #557, #568, #579, and #597 applied, PR #634 cherry-picked cleanly with zero manual conflict resolution.

## Verification

Local Go build is not possible without the sibling `sonic-mgmt-common` checkout; relying on the Azure pipeline in this PR for full build + test validation.


hdwhdw:
@vmittal-msft FYI

hdwhdw:
Force-pushed: dropped #568 (bypass) from the prerequisite chain. It was pulling in `pkg/bypass`, which transitively requires `github.com/redis/go-redis/v9` — a dependency that was bundled into #528 (mTLS Certz) on master and is absent on 202511. #568 isn't needed by #634 itself, so removing it eliminates the dependency gap without dragging in the gNSI mTLS stack.

Chain is now: #557 → #579 → #597 → #634.

hdwhdw:
@vmittal-msft As you requested here is the CP

hdwhdw:
Superceded by https://github.com/sonic-net/sonic-gnmi/pull/662

PR #661, opened 2026-04-30
[202511] Add OS.Verify and OS.Activate to DPU proxy forwardable methods

rookie-who:
Cherry-pick of #660 to the 202511 branch.

### Description
Add `OS.Verify` and `OS.Activate` to the DPU proxy forwardable methods registry so external clients can verify/activate OS versions on individual DPUs via the NPU proxy.

These RPCs are already implemented (`gnmi_server/gnoi_os.go`, `pkg/gnoi/os/os.go`) but were not registered in `defaultForwardableMethods`, causing requests with DPU metadata headers to be rejected.

See #660 for full context.


rookie-who:
Superseded by #662 which now includes forwarding handlers.

PR #671, opened 2026-05-06
gNSI:Add sonic service client support for Credentialz

niranjanivivek:
Signed-off-by: Neha Das nehadas@google.com

This PR enables support for managing credentials used for console and SSH access authentication on target device.

[HLD Link](https://github.com/rlucus/SONiC/blob/c4c043118d7cf07eab687dc4085f4af1f9346770/doc/mgmt/gnmi/gnsi.md)

Dependency Chain for Merge
Backend transformer PR : https://github.com/sonic-net/sonic-mgmt-common/pull/203
Backend sonic-host-services PR: https://github.com/sonic-net/sonic-host-services/pull/347
Frontend Base PR: [#671](https://github.com/sonic-net/sonic-gnmi/pull/671) - Merge first
Frontend Incremental PR : [#672](https://github.com/sonic-net/sonic-gnmi/pull/672)

### DIFF
--- a/sonic_service_client/dbus_client.go
+++ b/sonic_service_client/dbus_client.go
@@ -1,6 +1,7 @@
 package host_service
 
 import (
+	"context"
 	"fmt"
 	"reflect"
 	"strings"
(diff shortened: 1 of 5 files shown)


niranjanivivek:
@ndas7  Please check and merge

sonic_service_client/dbus_client.go
```diff
@@ -424,3 +454,96 @@ func (c *DbusClient) HealthzAck(req string) (string, error) {
...
+	if deadline, ok := ctx.Deadline(); ok {
+		remaining := time.Until(deadline)
+		if remaining <= 0 {
+			return context.DeadlineExceeded
+		}
+		timeout = int(remaining.Seconds())
+		if timeout > 10 {
+			timeout = 10
+		}
+	}
```
Bojun-Feng, line 546: Is there a typo here? Seems like it should be 300 in the nested if clause

niranjanivivek: HI @Bojun-Feng , Thanks for pointing this. 
Yes we think it is a typo error. So if timeout exceeds 300s, it can be reset. 
Is it fine to change the lines 543 -545 as below.? Please suggest if you have any comments here. 
if timeout > 300 {
        timeout = 300
}

niranjanivivek: Hi @ndas7 @Bojun-Feng . Please respond , so that I can update the code and submit for review.

Bojun-Feng: Hi @niranjanivivek, apologies for the delay. For some context, I noticed the discrepancy in the code since I usually paste code blocks and forget to update the value in nested conditions. However, I am not fully knowledgeable on the context. I believe @ndas7 needs to confirm whether we want to align the value to 10 or 300.

I would recommend a single source of truth in the code block cited for this comment. Define another variable named `threshold` / `timeout_cap` / something descriptive, then reuse it for the entire block so the hard coded number only appears one time. This should reduce similar issues in the future.

niranjanivivek: > Hi @niranjanivivek, apologies for the delay. For some context, I noticed the discrepancy in the code since I usually paste code blocks and forget to update the value in nested conditions. However, I am not fully knowledgeable on the context. I believe @ndas7 needs to confirm whether we want to align the value to 10 or 300.
> 
> I would recommend a single source of truth in the code block cited for this comment. Define another variable named `threshold` / `timeout_cap` / something descriptive, then reuse it for the entire block so the hard coded number only appears one time. This should reduce similar issues in the future.

@ndas7 Please suggest the correct timeout value here as requested by Bojun-Feng.

ndas7: The original PR https://github.com/sonic-net/sonic-gnmi/pull/573 was already merged - this PR was forked from that and is no longer needed. I agree there are some discrepancies here - timeout's initial value of 300 is arbitrary and can be reduced to 10. @niranjanivivek Could you close this PR and make the change in a standalone PR on top of the latest codebase? Thanks.

hdwhdw:
closing the PR for the author

PR #659, opened 2026-04-24
Revert "telemetry: add --bind_address flag to restrict TCP listener"

hdwhdw:
Reverts sonic-net/sonic-gnmi#652

PR #673, opened 2026-05-06
Revert "telemetry: add --bind_address flag to restrict TCP listener"

hdwhdw:
Reverts sonic-net/sonic-gnmi#652

Supersedes #659 — same revert plus a unit test (`TestCertAuthDisabledWhenNoCaCert`) covering the restored cert-auth fallback path so the coverage check passes.


yijingyan2 (approved):

yijingyan2:
Merging to unblock sonic-gnmi submodule update. To be investigated separately. @qiluo-msft for viz

PR #675, opened 2026-05-14
MixedDbClient: batch ZMQ Set/Del for multi-key gNMI updates (#27250)

hdwhdw:
### Why
`handleTableData` in `MixedDbClient` (under `sonic_data_client/mixed_db_client.go`) currently loops per key when servicing a gNMI Set:

- `opRemove` → calls `DbDelTable` per dbkey.
- multi-key JSON `opAdd` → calls `DbSetTable` per JSON top-level key.

For DASH route programming this is the hot path — a single gNMI Set can carry tens of thousands of keys, and we generate that many ZMQ round-trips even though they all target the same producer on the same DPU. `ZmqProducerStateTable` already has a batched `set(vector<KeyOpFieldsValuesTuple>)` / `del(vector<string>)` in `sonic-swss-common` that fires one ZMQ message regardless of N — but the Go bindings did not expose a callable form (see the dependency below).

### What
* New `DbBatchSetTable(table string, entries []tableBatchEntry) error` and `DbBatchDelTable(table string, keys []string) error` on `MixedDbClient`. For ZMQ-backed tables (`DASH_*` minus `DASH_HA_*`) they assemble parallel `VectorString` / `FieldValuePairsList` and call the new `swsscommon.ZmqProducerBatchedSet` / `swsscommon.ZmqProducerBatchedDel` helpers under `RetryHelper`. For plain `Table` and non-ZMQ `ProducerStateTable` backends — where there is no batched C++ API — they fall back to the existing per-entry loop, so behaviour for `DASH_HA_*` and ordinary CONFIG_DB writes is unchanged.
* `handleTableData`:
  - `opRemove` builds a key slice and makes a single `DbBatchDelTable` call.
  - multi-key JSON `opAdd` builds an entry slice and makes a single `DbBatchSetTable` call.
  - Single-key paths (1138, 1152) are untouched — batching one entry is pointless.
* `ZmqProducerBatchedSetWrapper` / `ZmqProducerBatchedDelWrapper` wrap the new SWIG entry points to convert C++ exceptions into Go errors (same pattern as the existing `ProducerStateTable…Wrapper`s).

### Tests
`sonic_data_client/client_test.go` gains:
* `TestZmqBatchedSetEndToEnd` / `TestZmqBatchedDelEndToEnd` — round-trip 5 entries through a real `ZmqProducerStateTable` + `ZmqConsumerStateTable` and verify the consumer drains them all.
* `TestDbBatchSetTableFallsBackForPlainTable` / `TestDbBatchSetTableEmptyAndSingle` — verify the `plainTableMap` fallback and the no-op/single-entry shortcuts.
* `TestHandleTableDataBatchesMultiKeyJSON` / `TestHandleTableDataBatchesOpRemove` — gomonkey-patched proof that `handleTableData` routes multi-key payloads through `DbBatchSetTable` / `DbBatchDelTable` exactly once (not N times).

### Related
* sonic-net/sonic-buildimage#27250 — tracking issue.
* **Depends on sonic-net/sonic-swss-common#1188** — exposes `ZmqProducerBatchedSet/Del/Send` to Go via SWIG `%inline` helpers. The `Set__SWIG_3(vector<KeyOpFieldsValuesTuple>)` overload that already exists in the generated `swsscommon.go` is unreachable from Go because the vector element type has no constructor (it's an opaque `Std_tuple_<…>`). PR #1188 adds parallel-vector shims that reuse the already-templated `VectorString` / `FieldValuePairsList` types.

**CI on this branch will fail to build until sonic-swss-common#1188 lands and a new `libswsscommon` artifact is published.** The compile errors will be on the `swsscommon.ZmqProducerBatched{Set,Del}` symbols.

### Out of scope (deliberate)
* Cross-update batching at the SetDB level — would change per-update atomicity semantics.
* Auto-chunking very large batches into multiple ZMQ messages — can be a follow-up if 100k+ in one message proves problematic on the orchagent side.
* `ZmqProducerBatchedSend` (mixed SET/DEL in one message) — exported by the swss-common PR for completeness but no gNMI caller needs it today.

### DIFF
--- a/sonic_data_client/mixed_db_client.go
+++ b/sonic_data_client/mixed_db_client.go
@@ -310,6 +310,18 @@ func ProducerStateTableDeleteWrapper(pt swsscommon.ProducerStateTable, key strin
 	return
 }
 
+func ZmqProducerBatchedSetWrapper(pt swsscommon.ZmqProducerStateTable, keys swsscommon.VectorString, fvss swsscommon.FieldValuePairsList) (err error) {
+	defer CatchException(&err)
+	swsscommon.ZmqProducerBatchedSet(pt, keys, fvss)
+	return
+}
+
+func ZmqProducerBatchedDelWrapper(pt swsscommon.ZmqProducerStateTable, keys swsscommon.VectorString) (err error) {
+	defer CatchException(&err)
+	swsscommon.ZmqProducerBatchedDel(pt, keys)
+	return
+}
+
 func TableSetWrapper(t swsscommon.Table, key string, value swsscommon.FieldValuePairs) (err error) {
 	// convert panic to error
 	defer CatchException(&err)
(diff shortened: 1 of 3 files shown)


prabhataravind:
@hdwhdw could you check PR failures and copilot comments?

sonic_data_client/mixed_db_client.go
```diff
@@ -396,6 +408,104 @@ func (c *MixedDbClient) DbDelTable(table string, key string) error {
...
+	values map[string]string
+}
+
+// DbBatchSetTable writes multiple entries to a single table. When the table
+// is backed by ZmqProducerStateTable it issues one ZMQ message for the
+// whole batch (via swsscommon.ZmqProducerBatchedSet), turning N round-trips
+// into one. For plain Table / non-ZMQ ProducerStateTable backends there is
+// no batched C++ API, so we fall back to a per-entry loop. Single-entry
+// batches are delegated to DbSetTable to avoid extra allocation.
+func (c *MixedDbClient) DbBatchSetTable(table string, entries []tableBatchEntry) error {
```
prabhataravind, line 424: Consider adding a log indicating the actual batch size being sent. This could be useful for debugging.

sonic_data_client/mixed_db_client.go
```diff
@@ -396,6 +408,104 @@ func (c *MixedDbClient) DbDelTable(table string, key string) error {
...
+			pair := swsscommon.NewFieldValuePair(k, v)
+			vec.Add(pair)
+			swsscommon.DeleteFieldValuePair(pair)
+		}
+		// Add() copies vec into the list; release the local copy.
+		fvss.Add(vec)
+		swsscommon.DeleteFieldValuePairs(vec)
+	}
+
+	return RetryHelper(
```
prabhataravind, line 467: At 125K+ entries, this single ZMQ message will exceed MQ_RESPONSE_MAX_COUNT (16MB). We will revisit that later based on tests.

prabhataravind:
@vivekrnv @croos12 for viz.

prabhataravind:
@qiluo-msft could you help review as well?

prabhataravind (dismissed):
One minor comment about logging the batch size. otherwise lgtm

sonic_data_client/mixed_db_client.go
hdwhdw: Done in a7db9d1 — added a `log.V(2).Infof` line on the ZMQ-backed path in both `DbBatchSetTable` and `DbBatchDelTable` printing the table name and batch size.

sonic_data_client/mixed_db_client.go
hdwhdw: Filed https://github.com/sonic-net/sonic-gnmi/issues/690 to track size-aware chunking in `DbBatchSetTable`/`DbBatchDelTable` so a single gNMI Set can exceed the 16 MiB ZMQ cap by splitting into multiple batched messages. Punting until we have measurements from real DASH bulk updates, as agreed.

prabhataravind (approved):
lgtm

PR #678, opened 2026-05-21
[gnoi] expand File.Put whitelist to /host/ + mkdir-p parent dirs

hdwhdw:
Two related changes to make `File.Put` usable for SONiC upgrade staging into the next-image overlay.

## 1) `validatePath` whitelist expansion (`/host/` allowed)

The current `allowedPrefixes = ["/tmp/", "/var/tmp/"]` blocks any write under `/host/`, including the legitimate next-image overlay staging path `/host/image-*/rw/etc/sonic/...`. The SONiC upgrade agent needs to put configs/certs into that directory between `InstallImage` and `Reboot`, so this PR appends `"/host/"` to the whitelist and updates the rejection error message accordingly.

The whitelist is **intentionally broad** for this PoC: it also accepts `/host/grub/grub.cfg` and `/host/machine.conf`. A follow-up PR will tighten the prefix to `/host/image-*/rw/` once callers stabilize. Mounted security is still enforced at the FS layer — without the companion sonic-buildimage PR ([gnmi] mount /host rw into gnmi container) the gnmi container only sees `/host` as read-only and writes return `EROFS` regardless of this whitelist.

The leading rationale comment block is rewritten to match the new behavior (the old block listed `/host/` as forbidden, which now contradicts the code).

### Residual risk (please review)

A client with `File.Put` access can now overwrite e.g. `/host/grub/grub.cfg` and brick the device on reboot. Mitigations: mTLS-gated RPC entry, the new mount in the buildimage PR is gnmi-only, and the follow-up will narrow the prefix. Happy to fold the narrower prefix into this PR if reviewers prefer; the upgrade agent's overlay path is fully covered by `/host/image-*/rw/`.

## 2) `HandlePut` mkdir-p parent dirs

`HandlePut` previously assumed the parent directory of the target path already existed; otherwise `os.OpenFile(tempPath, ...)` failed with `codes.Internal`. That forced every gNOI client to pre-create parent dirs via an out-of-band SSH mkdir, which defeats the purpose of `File.Put` as a self-contained upload primitive — especially painful for the upgrade agent because `/host/image-{next}/rw/etc/sonic/credentials/` usually does not exist yet on a freshly-installed next image.

This PR inserts `os.MkdirAll(filepath.Dir(tempPath), 0755)` right before the `os.OpenFile` call, after `validatePath` (Step 2) and `translatePathForContainer` (Step 3), so no new path can be reached that `validatePath` would not already accept.

## Tests

- `TestHandlePut_CreatesParentDir` asserts mkdir-p happens on a fresh deeply-nested `/tmp/<uniq>/sub/dir/file.txt` path.
- `TestHandlePut_ErrorPaths/"parent dir creation failure"` was rewritten to provoke an actual `MkdirAll` failure (a regular file planted where a parent dir would need to be), replacing the previous test which relied on the now-eliminated missing-parent failure path.
- Whitelist tests updated: `/host/grub/grub.cfg`, `/host/machine.conf` and `/host/image-master/rw/usr/bin/test` moved from the rejection table to the allowed-paths table (with an explicit `NOTE` about the broad prefix); a new positive case covers `/host/image-internal.164866913-.../rw/etc/sonic/minigraph.xml`; a `/host` traversal case (`/host/image-foo/../../etc/passwd`) verifies `filepath.Clean` resolves the traversal before the whitelist check.

`go test ./pkg/gnoi/file/...` is green locally.

## Companion PR

sonic-net/sonic-buildimage: `[gnmi] mount /host rw into gnmi container` — without it, this whitelist change has no effect inside the gnmi container.

### DIFF
--- a/pkg/gnoi/file/file.go
+++ b/pkg/gnoi/file/file.go
@@ -175,19 +175,21 @@ func translatePathForContainer(path string) string {
 // Allowed directories for SONiC devices:
 //   - /tmp/      - Temporary files, firmware images
 //   - /var/tmp/  - Temporary files that persist across reboots
+//   - /host/     - Next-image overlay staging (e.g. /host/image-*/rw/etc/sonic/...)
 //
 // Rejected paths include:
 //   - /etc/, /boot/, /usr/, /bin/, /sbin/ - Critical system directories
-//   - /host/ - Contains grub config, overlayfs layers, machine.conf
 //   - /var/log/ - System logs
 //   - /home/, /root/ - User home directories with SSH keys
 //   - Relative paths or paths with .. traversal
 //
-// Rationale: Only temporary directories are safe for firmware downloads.
-// Writing to /host/ risks:
-//   - Overwriting /host/grub/grub.cfg (brick device on reboot)
-//   - Corrupting /host/image-*/rw/ (overlayfs upperdir, kernel panic)
-//   - Modifying /host/machine.conf (platform detection failure)
+// Note on /host/: a broad /host/ prefix is currently accepted so the SONiC
+// upgrade agent can stage configs/certs into the next-image overlay at
+// /host/image-*/rw/etc/sonic/. This whitelist is intentionally permissive
+// for now; a follow-up will tighten the prefix to /host/image-*/rw/ once
+// callers stabilize. Writes under /host/ are still gated by filesystem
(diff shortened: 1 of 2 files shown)


ryanzhu706 (commented):
LGTM

ryanzhu706 (approved):

PR #679, opened 2026-05-21
[gnoi] SetPackage: auto-resolve version from image when not provided

ryanzhu706:
Description:

Problem

The System.SetPackage gNOI RPC requires the caller to provide the version field. If hwproxy or other callers don't provideit, the request fails with "version is missing in package request". The caller often doesn't know the version string ahead of time — it's embedded in the image binary itself.

Solution

Instead of failing when version is empty, the API now automatically resolves it by running sonic-installer binary_version <filename> after installation. This extracts the version from the image and uses it for sonic-installer set-default.

Behavior:

 - If version is provided → used directly (no change)
 - If version is empty and activate=true → auto-resolved via sonic-installer binary_version
 - If version is empty and activate=false → install-only succeeds without version

Tests

 - TestHandleSetPackage_ActivateWithAutoResolvedVersion — happy path
 - TestHandleSetPackage_AutoResolveVersionFails — error propagation
 - TestGetBinaryVersion_Success / _EmptyOutput / _CommandError — unit tests for helper

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/pkg/gnoi/system/system.go
+++ b/pkg/gnoi/system/system.go
@@ -86,10 +86,6 @@ func HandleSetPackage(ctx context.Context, req *syspb.SetPackageRequest) (*syspb
 		log.Errorf("Filename is missing in package request")
 		return nil, status.Errorf(codes.InvalidArgument, "filename is missing in package request")
 	}
-	if pkg.Package.Version == "" {
-		log.Errorf("Version is missing in package request")
-		return nil, status.Errorf(codes.InvalidArgument, "version is missing in package request")
-	}
 
 	// Reject RemoteDownload - require local image
 	if pkg.Package.RemoteDownload != nil {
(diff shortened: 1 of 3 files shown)


pkg/gnoi/system/system.go
hdwhdw, line 118: Revisiting the logic here and it all seems a bit messed up:
1. we should first validate input, meaning resolve the version before doing any write operation and reject it if necessary.
2. then we move to install
3. and then we move to activate.
Can we fix this?

hdwhdw (commented):
Can you include a manual test, possible with logs to make sure this work well? We can put it in PR description

hdwhdw (approved):

PR #683, opened 2026-05-22
Skip Redis client init for DBs not in database_config.json

yijingyan2:
#### Why I did it
`initRedisDbClients()` and `useRedisTcpClient()` iterate over every entry in the `spb.Target_value` proto enum and call `sdcfg.GetDbSock` / `sdcfg.GetDbTcpAddr` for each one. When a `Target` enum value corresponds to a DB that is not present in the device's database_config.json (e.g. DPU_COUNTERS_DB on a non-smartswitch / non-DPU platform), the underlying `SonicDBConfig::getDbInfo` call logs an ERR-level message to syslog:

 ```
ERR gnmi#telemetry: :- getDbInfo: Failed to find DPU_COUNTERS_DB database in : key
 ERR gnmi#dialout_client_cli: :- getDbInfo: Failed to find DPU_COUNTERS_DB database in : key
```

This was introduced as a side-effect of adding `DPU_COUNTERS_DB = 18` to the `Target` enum. Since `DPU_COUNTERS_DB` is only configured on smartswitch / DPU platforms, every other platform now emits these benign ERR logs every time `gnmi / dialout_client_cli` starts up.
#### How I did it
- Added a helper `configuredDbSet(dbNamespace)` that returns the set of DB names actually present in `database_config.json` for the given namespace.
 - In both `useRedisTcpClient()` (TCP path) and `initRedisDbClients()` (UNIX socket path), look up the configured DB set before iterating `spb.Target_value`. If a Target enum entry isn't in the configured set, skip it with a `V(2)` info log instead of calling `sdcfg.GetDbSock` / `sdcfg.GetDbTcpAddr` and triggering the ERR.
 - If `GetDbList` itself fails or returns empty, the helper returns nil and behaviour falls back to the original "try them all" path, so this is a strictly additive guard.
#### How to verify it
with this change, the test passed https://elastictest.org/scheduler/testplan/6a0ff4fe5ec6b5a1f3a20c7a?searchTestCase=rota&testcase=telemetry%2Ftest_telemetry_cert_rotation.py&type=console

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

### DIFF
--- a/sonic_data_client/client_test.go
+++ b/sonic_data_client/client_test.go
@@ -872,6 +872,53 @@ func TestInitRedisDbClients(t *testing.T) {
 		}
 	})
 
+	t.Run("SkipUnconfiguredDbWithoutGetDbSock", func(t *testing.T) {
+		defer saveAndResetTarget2RedisDb()()
+
+		const skippedDb = "CHASSIS_STATE_DB"
+		configuredDbs := []string{"CONFIG_DB", "APPL_DB", "STATE_DB"}
+		getDbSockCalls := 0
+		getDbSockCalledForSkipped := false
+
+		patches := gomonkey.ApplyFunc(sdcfg.GetDbAllNamespaces, func() ([]string, error) {
+			return []string{ns}, nil
+		})
+		defer patches.Reset()
+
+		patches.ApplyFunc(sdcfg.GetDbList, func(_ string) ([]string, error) {
+			return configuredDbs, nil
+		})
+
+		patches.ApplyFunc(sdcfg.GetDbSock, func(dbName string, _ string) (string, error) {
+			getDbSockCalls++
+			if dbName == skippedDb {
+				getDbSockCalledForSkipped = true
+			}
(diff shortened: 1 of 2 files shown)


theasianpianist (dismissed):
lgtm, suggest adding a new test case as per the copilot comment

theasianpianist (dismissed):

theasianpianist (approved):

hdwhdw:
@yijingyan2 already fixed by https://github.com/sonic-net/sonic-gnmi/commit/e8abbd63faa574cdbd9903e09fe6bf709b14f602?

yijingyan2:
> @yijingyan2 already fixed by [e8abbd6](https://github.com/sonic-net/sonic-gnmi/commit/e8abbd63faa574cdbd9903e09fe6bf709b14f602)?

yes, let me close the PR. Thanks

PR #705, opened 2026-06-16
Add v2r mapping for VoQ counters in VoQ Chassis

vmittal-msft:
#### Why I did it

VoQ chassis platforms expose VoQ counters in `COUNTERS_DB`, but the existing virtual DB translation logic does not support querying them through gNMI using switch-level or system-port-level paths. As a result, VoQ statistics are not available through the same translated interface that already exists for other counter types.

This change adds the missing v2r mapping and namespace-aware counter resolution needed to expose VoQ counters for VoQ chassis deployments.

#### How I did it

- Added `v2rSystemPortVoQStats()` to translate VoQ counter queries from virtual DB paths to the underlying `COUNTERS_DB` OID-backed entries.
- Registered the `*/VoQs` path in `pathTransFuncTbl` so VoQ stats can be queried using:
  - wildcard switch queries
  - single-switch queries
  - system-port-specific queries
- Added helper functions:
  - `parseVoQName()` to parse VoQ names in `switch|asic|port:voq` format
  - `buildVoQJsonKey()` to build the translated JSON response keys
- Added `initCountersVoQNameMap()` and `getVoQCountersMap()` to initialize the VoQ name-to-OID mapping from `COUNTERS_VOQ_NAME_MAP`
- Added namespace-aware OID resolution so the implementation can identify which Redis namespace actually owns the VoQ counter OID data on multi-ASIC / VoQ chassis systems
- Added `resolveVoQNamespace()` to safely use cached namespace mappings and fall back to the namespace derived from the VoQ name when needed
- Fixed trie wildcard matching in `findNode()` to support backtracking, so real switch names correctly resolve to the new `*/VoQs` path without breaking existing `Ethernet*` path handling
- Added unit and testdata coverage for:
  - wildcard VoQ queries
  - single-switch VoQ queries
  - system-port VoQ queries
  - namespace resolution behavior
  - invalid VoQ name handling
  - trie backtracking behavior

#### How to verify it

- Run the unit tests covering the new VoQ translation logic and trie behavior
- Verify that gNMI queries against `COUNTERS_DB` return translated VoQ counter data for:
  - `COUNTERS/SwitchName*/VoQs`
  - `COUNTERS/<switch-name>/VoQs`
  - `COUNTERS/<switch-name>|<asic>|<system-port>/VoQs`
- Confirm that returned data is resolved from the correct Redis namespace on VoQ chassis / multi-ASIC platforms
- Confirm that existing counter path translation for Ethernet and queue paths continues to work as before
- Verify the newly added test coverage passes with the included VoQ testdata and OID fixtures
- Please find test results in comment section (file : pr705_testing_done.md)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111


vmittal-msft:
Test transcript with output : 
[pr705_testing_done.md](https://github.com/user-attachments/files/29058681/pr705_testing_done.md)

sonic_data_client/virtual_db.go
```diff
@@ -1258,6 +1338,159 @@ func InitCountersSidMap() error            { return initCountersSidMap() }
...
+// cached namespace is currently active (part of the active namespace set returned
+// by GetRedisClientsForDb), and falls back to deriving the namespace from the
+// VoQ name's asic component.
+func resolveVoQNamespace(oid, asicNamespace string, activeNamespaces map[string]*redis.Client) string {
+	if ns, ok := countersVoQOidNamespaceMap[oid]; ok {
+		if activeNamespaces[ns] != nil {
+			return ns
+		}
+	}
+	return strings.ToLower(asicNamespace)
```
hdwhdw: When discovery misses, guessing the namespace from the name can read the wrong Redis (silent empty result) or, on a chassis, name a non-local namespace and nil-deref in the reader. Consider not guessing, and skipping the VoQ instead:

```go
func resolveVoQNamespace(oid, asicNamespace string, active map[string]*redis.Client) (string, bool) {
    if ns, ok := countersVoQOidNamespaceMap[oid]; ok && active[ns] != nil {
        return ns, true
    }
    return "", false // unknown owner; caller skips this VoQ rather than guess
}
```

sonic_data_client/virtual_db.go
```diff
@@ -1258,6 +1338,159 @@ func InitCountersSidMap() error            { return initCountersSidMap() }
...
+func v2rSystemPortVoQStats(paths []string) ([]tablePath, error) {
+	// paths[DbIdx] = "COUNTERS_DB"
+	separator, _ := GetTableKeySeparator(paths[DbIdx], "")
+	// Get the currently active Redis clients for namespace validation
+	activeNamespaces, _ := GetRedisClientsForDb(paths[DbIdx])
+	var tblPaths []tablePath
+
+	// VoQs on all SwitchName* (str2-7804*)
+	if strings.HasSuffix(paths[KeyIdx], "*") {
+		for voqName, oid := range countersVoQNameMap {
```
hdwhdw, line 1409: This ranges the shared maps without the lock that `ClearMappings()` uses, so it can race. Consider guarding the body like `v2rEthPortQueStats` does:

```go
clearMappingsMu.RLock()
defer clearMappingsMu.RUnlock()
```

sonic_data_client/virtual_db.go
```diff
@@ -234,6 +244,74 @@ func initCountersAclRuleMap() error {
...
+			log.V(2).Infof("redis HGetAll failed for COUNTERS_DB in namespace %v, tableName: %s: %v", namespace, tableName, err)
+			return nil, nil, err
+		}
+		addmap(counterMap, fv)
+		log.V(6).Infof("tableName: %s in namespace %v, map entries: %d", tableName, namespace, len(fv))
+	}
+
+	// Second pass: determine which namespace has counter data for each OID
+	for namespace, redisDb := range redisClientMap {
+		keys, err := redisDb.Keys(context.Background(), "COUNTERS:oid:*").Result()
```
hdwhdw: `KEYS` scans the whole keyspace and blocks the Redis instance while it runs. Consider `SCAN`:

```go
iter := redisDb.Scan(context.Background(), 0, "COUNTERS:oid:*", 0).Iterator()
for iter.Next(context.Background()) {
    existingOids[strings.TrimPrefix(iter.Val(), "COUNTERS:")] = true
}
if err := iter.Err(); err != nil {
    log.V(2).Infof("scan failed in namespace %v: %v", namespace, err)
    continue
}
```

sonic_data_client/virtual_db.go
```diff
@@ -234,6 +244,74 @@ func initCountersAclRuleMap() error {
 	return initErr
 }
 
+func initCountersVoQNameMap() error {
+	clearMappingsMu.RLock()
+	alreadyInit := len(countersVoQNameMap) != 0
```
hdwhdw, line 249: This oid→namespace map is built once and never refreshed, so if a counter moves namespace (warm reboot, LC insert/remove) it routes to the stale one forever. Consider documenting that ownership is assumed fixed for the process lifetime, or wiring an invalidation hook so this can be rebuilt.

hdwhdw (approved):

PR #710, opened 2026-06-25
telemetry: default --bind_address to loopback when --noTLS has none

hdwhdw:
#### Why I did it

#709 (re-land of #652) made `setupFlags()` **reject** `--noTLS` unless `--bind_address` is an explicit loopback address. But the default SONiC start path (`dockers/docker-sonic-gnmi/gnmi-native.sh`) appends `--noTLS` with **no** `--bind_address` whenever no telemetry certs are configured in ConfigDB — which is the case on a stock image (KVM/vs, and any box without certs). As a result `/usr/sbin/telemetry` exits immediately at startup and `gnmi-native` never runs:

```
ERR gnmi#supervisor-proc-exit-listener-rs: Process 'gnmi-native' is not running in namespace 'host'
```

This regresses the whole gNMI/telemetry test cluster (`telemetry/test_telemetry`, `gnmi/test_gnmi_appldb`, `gnmi/test_gnmi_smartswitch`, `zmq/test_gnmi_zmq`, `autorestart/test_container_autorestart`) the moment the submodule is bumped (e.g. sonic-buildimage PR #28084), because the server now requires explicit config just to start.

#### How I did it

Make the `--noTLS` path **secure-by-default instead of crash-by-default** in `telemetry/telemetry.go`:

- When `--noTLS` is set and `--bind_address` is **empty**, default it to `127.0.0.1` (loopback) and log a warning — the server starts and cleartext gRPC is confined to localhost.
- When `--noTLS` is set with an **explicit non-loopback** address, still return an error — the original cleartext-exposure guard is preserved for the genuinely unsafe misconfiguration (`--noTLS --bind_address 0.0.0.0`).

`gnmi_server/server.go` already handles an empty/loopback `BindAddress` correctly, so no change is needed there.

#### How to verify it

- `TestNoTLSDefaultsToLoopback` (new): `--noTLS` with no `--bind_address` succeeds and resolves `BindAddress` to `127.0.0.1` on both `TelemetryConfig` and `gnmi.Config`.
- `TestNoTLSRequiresLoopbackAddress` (updated): empty `--bind_address` now expects success (defaulted), non-loopback still errors, loopback v4/v6 succeed.
- End-to-end: a sonic-buildimage submodule bump to this commit should bring `gnmi-native` up on a stock image and clear the telemetry/gnmi LogAnalyzer failures.

telemetry: default `--bind_address` to loopback when `--noTLS` is set without an address, so the server starts under the default (no-cert) configuration while still rejecting cleartext gRPC on a routable address.

### DIFF
--- a/telemetry/telemetry.go
+++ b/telemetry/telemetry.go
@@ -237,8 +237,11 @@ func setupFlags(fs *flag.FlagSet) (*TelemetryConfig, *gnmi.Config, error) {
 	}
 
 	if *telemetryCfg.NoTLS {
-		ip := net.ParseIP(*telemetryCfg.BindAddress)
-		if ip == nil || !ip.IsLoopback() {
+		if *telemetryCfg.BindAddress == "" {
+			// Default the cleartext listener to loopback.
+			*telemetryCfg.BindAddress = "127.0.0.1"
+			log.Warning("--noTLS set without --bind_address; defaulting to 127.0.0.1 (loopback only)")
+		} else if ip := net.ParseIP(*telemetryCfg.BindAddress); ip == nil || !ip.IsLoopback() {
 			return nil, nil, fmt.Errorf(
 				"--noTLS requires --bind_address to be a loopback address (e.g. 127.0.0.1 or ::1) " +
 					"to prevent cleartext gRPC exposure over the network")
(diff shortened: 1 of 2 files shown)


yijingyan2:
Hi @hdwhdw, sorry for the late review, could you please help resolve the code conflict? Thanks

hdwhdw:
@yijingyan2 should have closed this. The solution is superceded https://github.com/sonic-net/sonic-gnmi/commit/46f96f841dc63fc245ce3f01804a8ac49a117041

PR #711, opened 2026-06-25
telemetry: remove --noTLS loopback address enforcement

xq9mend:
The --bind_address flag added in #709 is useful optional hardening, but the enforcement that requires --bind_address to be a loopback address when --noTLS is active is too strict.

Operators may legitimately use --noTLS without --bind_address in isolated networks, container environments, or lab setups. Enforcing loopback-only breaks backward compatibility.

This PR removes the enforcement block and the test that covered it. The --bind_address flag itself is retained -- operators can still restrict the listener to localhost when desired.

### DIFF
--- a/telemetry/telemetry.go
+++ b/telemetry/telemetry.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"net"
 	"os"
 	"os/signal"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


xq9mend:
@hdwhdw @zbud-msft — could you please review this PR? It removes the loopback address enforcement that was added to the `--noTLS` code path. The enforcement blocked legitimate remote gNMI connections (e.g. from test infrastructure). This is a follow-up cleanup to #709.

hdwhdw (approved):

PR #733, opened 2026-08-04
COUNTERS_DB bare-table enumeration and BUFFER_POOL_WATERMARKS virtual path support

Verma-Anukul:
## Summary
Fix bare-path Get/Subscribe for keyed COUNTERS_DB tables (notably `BUFFER_POOL_WATERMARKS`) and strip SONiC NULL placeholders from JSON.

## Before

**gNMI Get** on bare path `/COUNTERS_DB/BUFFER_POOL_WATERMARKS` returned empty or incomplete results (pools required an explicit wildcard).

**gNMI JSON output** for hashes carrying only a SONiC placeholder:
```json
{"NULL": "NULL"}
```

**Subscribe / ONCE mode** (`gnmic --mode once`) on keyed COUNTERS_DB tables could hang or miss updates because PSUBSCRIBE patterns did not match keyed rows like `BUFFER_POOL_WATERMARKS:<oid>`.

**gnmi.log** (when key enumeration failed):
```
listBareTableKeys failed for {dbName:COUNTERS_DB tableName:BUFFER_POOL_WATERMARKS ...}: ...
```

## After

**gNMI Get** on `/COUNTERS_DB/BUFFER_POOL_WATERMARKS` returns all buffer pools by name (via `COUNTERS_BUFFER_POOL_NAME_MAP` → `COUNTERS:<oid>`).

**gNMI JSON output** — NULL placeholder stripped:
```json
{}
```
or with real fields:
```json
{"ingress_lossless_pool": {"SAI_BUFFER_POOL_STAT_WATERMARK_BYTES": "1234"}}
```

**Subscribe** uses corrected keyspace patterns with `delimSkipped` for COUNTERS_DB non-`COUNTERS` tables; ONCE-mode subscriptions exit cleanly.

## Test plan
- [ ] `go test ./sonic_data_client/... -run TestMakeJSONRedisSkipsNull`
- [ ] gNMI Get on `/COUNTERS_DB/BUFFER_POOL_WATERMARKS` → all pools
- [ ] `gnmic --mode once` subscribe on buffer pool watermarks → exits cleanly

### DIFF
--- a/sonic_data_client/virtual_db.go
+++ b/sonic_data_client/virtual_db.go
@@ -193,6 +193,7 @@ func initCountersPGNameMap() error {
 			initErr = err
 			return
 		}
+		local := make(map[string]map[string]string)
 		for pg, oid := range pgOidMap {
 			// pg is in format of "Ethernet64:7"
 			pg_parts := strings.Split(pg, ":")
@@ -200,11 +201,12 @@ func initCountersPGNameMap() error {
 				initErr = fmt.Errorf("invalid pg name %v", pg)
 				return
 			}
-			if _, ok := countersPGNameMap[pg_parts[0]]; !ok {
-				countersPGNameMap[pg_parts[0]] = make(map[string]string)
+			if _, ok := local[pg_parts[0]]; !ok {
+				local[pg_parts[0]] = make(map[string]string)
 			}
-			countersPGNameMap[pg_parts[0]][pg_parts[1]] = oid
+			local[pg_parts[0]][pg_parts[1]] = oid
 		}
+		countersPGNameMap = local
 	})
 	return initErr
 }
@@ -389,7 +391,7 @@ func initCountersBufferPoolNameMap() error {
 	if err != nil {
 		return err
 	}
-	countersBufferPoolNameByNamespace = make(map[string]map[string]string)
+	local := make(map[string]map[string]string)
 	for namespace, redisDb := range redis_client_map {
 		_, err := redisDb.Ping().Result()
 		if err != nil {
@@ -408,8 +410,9 @@ func initCountersBufferPoolNameMap() error {
 		for k, v := range fv {
 			poolMap[k] = v
 		}
-		countersBufferPoolNameByNamespace[namespace] = poolMap
+		local[namespace] = poolMap
 	}
+	countersBufferPoolNameByNamespace = local
 	return nil
 }
 
--- a/sonic_data_client/virtual_db.go
+++ b/sonic_data_client/virtual_db.go
@@ -85,6 +85,9 @@ var (
 	// Mutex to protect ClearMappings from racing with init functions
 	clearMappingsMu sync.RWMutex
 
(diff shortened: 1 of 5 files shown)


sonic_data_client/virtual_db.go
```diff
@@ -135,6 +138,12 @@ var (
 		}, { // specific field stats for PORT_PHY_ATTR for one or all Ethernet ports (no alias translation)
 			path:      []string{"COUNTERS_DB", "PORT_PHY_ATTR", "Ethernet*", "*"},
 			transFunc: v2rTranslate(v2rPortPhyAttrFieldStats),
+		}, { // Buffer pool watermarks - bare path returns all pools
+			path:      []string{"COUNTERS_DB", "BUFFER_POOL_WATERMARKS"},
```
hdwhdw, line 142: This virtual mapping is applied only through `DbClient`. The native `sonic-db` route constructs `MixedDbClient`, whose path resolver does not call `lookupV2R`; `/sonic-db:COUNTERS_DB/localhost/BUFFER_POOL_WATERMARKS` therefore searches raw `BUFFER_POOL_WATERMARKS:*` rows rather than `COUNTERS:<oid>` and cannot return named buffer pools. Apply the translation in the native path, or explicitly limit the feature; add a native `sonic-db` Get test that asserts pool-name-keyed output.

hdwhdw (commented):
Two inline findings on API coverage and mapping lifecycle.

sonic_data_client/virtual_db.go
```diff
@@ -370,6 +381,51 @@ func initDebugNameSwitchStatMap() error {
 	return nil
 }
 
+func initCountersBufferPoolNameMap() error {
+	clearMappingsMu.Lock()
+	defer clearMappingsMu.Unlock()
+
+	value := os.Getenv("UNIT_TEST")
+	if len(countersBufferPoolNameByNamespace) > 0 && value != "1" {
```
hdwhdw: This early return makes the pool-name map a process-lifetime snapshot. A pool added after the first request remains invisible, and a deleted pool can still resolve to its stale OID. Refresh or invalidate this mapping when `COUNTERS_BUFFER_POOL_NAME_MAP` changes, and add tests for pools created and deleted after the initial request.

hdwhdw (commented):
Inline finding on buffer-pool mapping lifecycle.

sonic_data_client/virtual_db.go
Verma-Anukul: lookupV2R() is now applied in MixedDbClient.getDbtablePath() for COUNTERS_DB, and jsonTableKey is honored in tableData2Msi(). TestMixedDbClientGetBufferPoolWatermarks asserts pool-name-keyed Get output on the native path.

sonic_data_client/virtual_db.go
Verma-Anukul: Removed the process-lifetime early return in initCountersBufferPoolNameMap() so the map is refreshed from Redis on each call. TestInitCountersBufferPoolNameMap_RefreshesOnChange covers add/delete after the first request.

PR #721, opened 2026-07-22
telemetry: resolve no-TLS link-local bind by interface

hdwhdw:
#### Why I did it

SmartSwitch DPUs expose plaintext gNMI/gNOI only on their chassis-internal
IPv4 link-local midplane. The NPU proxy must reach that address, but the current
`--noTLS` validation permits loopback addresses only. This makes it impossible
to restore proxy connectivity without removing the cleartext bind protection
entirely.

This is the binary-side dependency for
https://github.com/sonic-net/sonic-buildimage/issues/28540.

#### How I did it

- Added an explicit, default-off `--no_tls_link_local_interface` option.
- Kept loopback as the default requirement for `--noTLS`.
- Resolve exactly one IPv4 link-local address from the named interface and bind
  that concrete address.
- Wait up to 30 seconds for address readiness, then fail nonzero if the address
  remains unavailable or ambiguous.
- Reject simultaneous explicit bind-address configuration and use with a
  non-default gNMI VRF.
- Added deterministic tests for arguments, delayed readiness, timeout, and
  zero/one/multiple link-local address selection.

#### How to verify it

- Native arm64 SONiC package and container build succeeded, including gofmt
  validation and compilation of the telemetry binary.
- On a physical SmartSwitch DPU, the built binary resolved the concrete
  `eth0-midplane` IPv4 link-local address itself.
- With the dependent launch-script change, both DPU services bound only to the
  concrete midplane link-local address. Direct gNOI `System.Time` and the real
  NPU UDS metadata-routed DPUProxy call succeeded.
- Negative TCP probes confirmed the services were not listening on the DPU
  management address or loopback.

#### Which release branch to backport (provide reason below if selected)

- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511
- [ ] 202605

Tracking issue/work item for backport/cherry-pick request (GitHub issue or Microsoft ADO): N/A
Failure type: regression

#### Tested branch

- [x] master
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511
- [ ] 202605
- [ ] N/A

#### Test result

- master: native arm64 build passed; physical SmartSwitch direct and proxied
  gNOI validation passed.

#### Description for the changelog

Allow explicitly selected plaintext gNMI binding to an interface's IPv4 link-local address.

#### Link to config_db schema for YANG module changes

N/A

PR #749, opened 2026-08-07
dpuproxy: use TLS for DPU connections

hdwhdw:
## Why

Certificate-free DPU telemetry servers use ephemeral self-signed TLS. The NPU DPU proxy currently uses plaintext gRPC credentials, so it cannot connect when the server runs with `--insecure`.

Tracks sonic-net/sonic-buildimage#28540. Replaces #721.

Microsoft ADO (number only): 39179390

## What

- Use TLS 1.2+ for DPU proxy connections and skip verification of the ephemeral server certificate.
- Add a real self-signed TLS `System.Time` connection test with optional client-certificate request behavior.
- Record certificate CN as an audit principal only when the TLS chain was verified; unverified optional certificates cannot spoof principal identity.

## Validation

- `go test -gcflags=all=-l ./pkg/interceptors/dpuproxy` - passed
- `go test ./pkg/interceptors -run TestDeriveRequestFields -count=1` - passed
- Self-signed TLS connection test passed repeatedly

## Remaining Validation

- Final rebuilt integrated image validation remains pending.


hdwhdw:
Separate targeted and full sonic-gnmi CI passed. Physical validation used Azure image run `1187671` on a one-NPU/four-BF3 SmartSwitch: the packaged DPU0 launcher served TLS, direct `gnoi_client -insecure` `System.Time` passed 5/5, and `System.Time` passed through DPUProxy on the physical NPU using metadata-routed UDS.

This TLS mode encrypts isolated-midplane traffic without certificate-based peer identity verification. CA-backed mTLS and explicit password/JWT authentication are available, and an unverified client CN is not logged as the principal.

Run `1187671` remains in progress overall; its BF3 and Mellanox jobs and artifacts passed.

qiluo-msft (approved):

PR #769, opened 2026-08-25
test: make fixture dependencies configurable

hdwhdw:
## Why

Several test fixtures use these fixed production paths:

- `/var/log`
- `/mnt/host`
- `/etc/sonic`
- `/var/run/redis/sonic-db`

Using these paths requires privileged setup. Fixtures can also write outside their workspaces.

The fixtures use the default System V shared-memory key `7749`. Test runs that use this key can share state.

A separate fixture migration will create a runtime directory tree owned by the tests. The migration will use an isolated shared-memory key and remove `sudo` where privileged access is not required.

This pull request adds only the inputs, validation, and unit tests required by that migration. It does not relocate fixtures or remove `sudo`.

## Changes

- If `SONIC_GNMI_SHM_KEY` is set, its value must be a base-10 integer from `1` through `2147483647`.
- Package-local variables define the journal and Healthz root paths.
- If `SONIC_GNMI_CHECKPOINT_DIR` is set, its value must be an absolute path.
- The service rejects a relative checkpoint path before any D-Bus or database access.
- `swsscommon` uses `DB_CONFIG_PATH` during initialization.
- Journal rotation creates and deletes backups in the directory selected when the journal is created.
- Focused tests cover the default, override, validation, caller-path, and rotation behavior.

## Scope

The existing defaults apply when override variables are unset. This pull request does not change fixture paths, the System V interprocess communication namespace, or checkpoint mounts.

The fixture migration must give telemetry and `gnmi_dump` the same shared-memory key. It must mount a custom checkpoint directory at the same absolute path for telemetry and the host service.

## Validation

These commands passed in a SONiC `sonic-slave-trixie` environment:

```sh
go test -race -mod=vendor ./common_utils -run "^TestSharedMemoryKey$" -count=1
go test -mod=vendor ./sonic_db_config -run "^TestDbConfigFiles$" -count=1
go test -race -mod=vendor ./sonic_data_client -run "^(TestCheckPointPath|TestCheckpointPathCallers|TestCheckpointPathCallersRejectRelativePath)$" -count=1
go test -race -mod=vendor ./gnmi_server -run "^(TestDbJournalPath|TestHealthzArtifactPath|TestDbJournalRotationUsesFileDirectory|TestNewServerRejectsInvalidSharedMemoryKey)$" -count=1
go test -race -mod=vendor ./telemetry -run "^TestRunTelemetryRejectsInvalidSharedMemoryKey$" -count=1
```

Tracking: Azure DevOps Task [39404784](https://msazure.visualstudio.com/One/_workitems/edit/39404784).


hdwhdw:
Focused dependency-selector tests passed for `common_utils`, `sonic_db_config`, `sonic_data_client`, and `gnmi_server`.

The tests ran in a temporary `sonic-slave-trixie` container. The test environment used the repository's Go 1.25.9 toolchain, generated `swsscommon` bindings, patched vendor sources, and local AMD64 development packages for SONiC. The container used a copy of the source tree from pull request head commit `e0e959c8dc2e0fd6d2042470306081d035b92cf3`. The pull request changes do not include generated files or dependency changes.

The tests validate default values for five selectors, injected values, and handling of invalid shared-memory keys. These results do not validate fixture migration, rootless test execution, end-to-end system services, the full package suite, or continuous integration suites.

sonic_data_client/mixed_db_client.go
```diff
@@ -47,6 +47,14 @@ const MAX_RETRY_COUNT uint = 5
 const RETRY_DELAY_MILLISECOND uint = 100
 const RETRY_DELAY_FACTOR uint = 2
 const CHECK_POINT_PATH string = "/etc/sonic"
+
+func checkPointPath() string {
+	if path := os.Getenv("SONIC_GNMI_CHECKPOINT_DIR"); path != "" {
```
donghaolicd, line 53: Please reject relative values here, or define separate host-visible and container-visible paths. This value is passed over D-Bus to the host-side `config checkpoint` command and is then opened by telemetry in its own filesystem. For example, `tmp/checkpoint` is resolved under the host checkpoint directory by GCU, while telemetry reads it relative to its working directory, so it cannot read the checkpoint it requested. The new test uses an absolute `t.TempDir()` and pre-creates the local file, so the fake D-Bus service does not cover this mismatch.

common_utils/shareMem.go
```diff
@@ -18,8 +20,24 @@ var (
...
+	if err != nil || parsed <= 0 {
+		return 0, fmt.Errorf("invalid SONIC_GNMI_SHM_KEY %q", value)
+	}
+	return int(parsed), nil
+}
+
 func SetMemCounters(counters *[int(COUNTER_SIZE)]uint64) error {
-	shmid, _, err := syscall.Syscall(syscall.SYS_SHMGET, uintptr(memKey), uintptr(memSize), uintptr(memMode))
+	key, err := sharedMemoryKey()
+	if err != nil {
```
donghaolicd, line 43: Could we propagate this error to server startup? `NewServer` calls `common_utils.InitCounters()`, but `InitCounters` discards the error from `SetMemCounters` (and `IncCounter` does the same). As a result, `SONIC_GNMI_SHM_KEY=invalid` does not reject the configuration: telemetry still starts and counter publication silently stops. Please return/handle the initialization error and add a server-level test for an invalid key.

gnmi_server/db_journal.go
```diff
@@ -247,7 +253,7 @@ func (dbj *DbJournal) rotateFile() error {
 		}
 
 		// Remove a rotated, zipped file if the maxBackups limit is reached
-		files, err := os.ReadDir(HostVarLogPath)
+		files, err := os.ReadDir(hostVarLogPath)
```
donghaolicd: Could rotation derive its directory from `filepath.Dir(dbj.fileName)` instead of rereading the mutable package global? `journal.fileName` is fixed when the journal is created, but rotation uses the current `hostVarLogPath` for scanning, deleting, and creating backups. If a test restores that global while the journal goroutine is still alive, the active file can remain in directory A while rotation operates in directory B, potentially deleting an unrelated backup. A test that changes the global after constructing the journal would cover this lifecycle.

common_utils/shareMem.go
hdwhdw: Commit `56627c3` makes `NewServer` reject a shared-memory key that fails validation. `runTelemetry` validates the key before starting any signal or server worker. This validation prevents `runTelemetry` from blocking if `NewServer` rejects the key. The service does not guarantee a successful counter write during a System V shared-memory failure.

sonic_data_client/mixed_db_client.go
hdwhdw: Commit `56627c3` requires `SONIC_GNMI_CHECKPOINT_DIR` to contain an absolute path. `SetIncrementalConfig`, `GetCheckPoint`, and the CONFIG_DB `Get` fallback return the validation error before any D-Bus or database access. Tests assert the exact error. Tests verify that no D-Bus or database access occurs.

gnmi_server/db_journal.go
hdwhdw: Commit `56627c3` changes journal rotation to use `filepath.Dir(dbj.fileName)`. After journal creation, the test changes `hostVarLogPath`. The test verifies backup creation and deletion of the oldest backup in the journal directory. The test verifies that a backup under the updated `hostVarLogPath` remains unchanged.

donghaolicd (approved):

